PR #4, opened 2016-05-13
WIP: Register a custom type

tjwp:
This is a very rough prototype for supporting custom types in generate models. The basic idea is to allow values to be stored in the model that require conversion to/from avro.

A `.from_avro` method (if defined on the value class) or a supplied proc is used as a coercer when values are assigned to the model.

A `.to_avro` method (if defined on the value class) or a supplied proc is used to prepare values before they are serialized using Avro.

An example configuration to use this is:

```ruby
  config.register_type('com.salsify.salsify_uuid', SalsifyUuid) do |type|
    type.from_avro = ->(value) { SalsifyUuid.parse(value) }
    type.to_avro = ->(value) { value.to_s }
  end
end
```

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -81,6 +81,12 @@ module Avromatic
         end
 
         def avro_field_class(field_type)
+          if Avromatic.custom_types.key?(field_type.try(:fullname))
+            custom_type = Avromatic.custom_types[field_type.fullname]
+            value_class = custom_type.value_class
+            return value_class if value_class
+          end
+
           case field_type.type_sym
           when :string, :bytes, :fixed
             String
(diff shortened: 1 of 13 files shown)


lib/avromatic/model/serialization.rb
```diff
@@ -41,7 +41,7 @@ def avro_hash(fields)
             result[key.to_s] = if value.is_a?(Avromatic::Model::Attributes)
                                  value.value_attributes_for_avro
                                else
-                                 value
+                                 avro_coder[key].try(:call, value) || value
```
jturkel: Should we give the `avro_coder` a default no-op coder so we can avoid the `try`?

jturkel: Also is this looking up the coder based on the name of the field rather than the type?

lib/avromatic/model/attributes.rb
```diff
@@ -81,6 +81,12 @@ def required?(field)
         end
 
         def avro_field_class(field_type)
+          if Avromatic.custom_types.key?(field_type.try(:fullname))
+            custom_type = Avromatic.custom_types[field_type.fullname]
+            value_class = custom_type.value_class
+            return value_class if value_class
```
jturkel: When would `value_class` be falsey?

lib/avromatic/model/serialization.rb
tjwp: I updated this to stop using `try` because I don't want to assume `ActiveSupport`.

tjwp: The fullname of the type is not readily accessible at this point. We could prebuild a map from the field names to the Avro fields. The we could lookup name -> Avro field -> custom type.

Here, I just prebuilt the map from field name to coder.

lib/avromatic/model/attributes.rb
tjwp: The way this is currently implemented, a value class is not required. You can specify a "custom type" based on an Avro fullname, then associate `to_avro`/`from_avro` Procs with that. The class associated with the Virtus attribute will then be based on the builtin Avro type.

jturkel:
Overall I think the approach looks great! We should probably go ahead and pull `SalsifyUuid` out of `activerecord-extensions` so we can setup this configuration without any unnatural dependencies.

tjwp:
Real PR here: https://github.com/salsify/avromatic/pull/6

PR #6, opened 2016-05-17
Add support for custom types

tjwp:
This change allows custom types to be registered for named types (record, enum, fixed). As part of the configuration for a custom type, a class can be specified for attribute values and methods for conversion to and from Avro.

Further details in the README.

Prime: @jturkel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 - Allow the specification of a custom type, including conversion to/from Avro,
   for named types.
 
+## v0.3.0
+- Remove dependency on the `private_attr` gem.
+
 ## v0.2.0
 - Allow a module level schema registry to be configured.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # avromatic changelog
 
-## v0.3.0
-- Remove dependency on the `private_attr` gem.
+## v0.4.0
+- Allow the specification of a custom type, including conversion to/from Avro,
+  for named types.
 
 ## v0.2.0
 - Allow a module level schema registry to be configured.
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 
 [travis]: http://travis-ci.org/salsify/avromatic
 
-`Avromatic` generates Ruby models from Avro schemas and provides utilities to
-encode and decode them.
+`Avromatic` generates Ruby models from [Avro](http://avro.apache.org/) schemas
+and provides utilities to encode and decode them.
 
 ## Installation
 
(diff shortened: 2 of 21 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # avromatic changelog
 
-## v0.3.0
```
jturkel: Was this changelog entry intentionally removed?

tjwp: Bad merge. I'll restore it.

README.md
```diff
@@ -88,6 +97,40 @@ constant:
...
 ```
 
+#### Custom Types
+
+Custom types can be configured for fields of named types (record, enum, fixed).
+These customizations are registered on the `Avromatic` module. Once a custom type
+is registered, it is used for all models with a schema that references that type.
+
+```ruby
+Avromatic.register_type('com.example.my_string', MyString)
```
jturkel: Should we recommend that people do this in an `Avromatic.configure` block?

lib/avromatic/model/attributes.rb
```diff
@@ -9,6 +9,15 @@ module Model
     module Attributes
       extend ActiveSupport::Concern
 
+      def self.first_union_schema(field_type)
```
jturkel: firstunionlogo svg

lib/avromatic/model/custom_type.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def initialize(value_class)
+        @value_class = value_class
+      end
+
+      # A coercer method is used when assigning to the model. It is used both when
+      # deserializing a model instance from Avro and when directly instantiating
+      # an instance. The coecer method must accept a single argument and return
+      # the value to store in the model for the attribute.
+      def coercer
```
jturkel: Should we call this `decoder` (or better yet `avro_decoder` and `avro_encoder` instead of `coder`)? It wasn't clear to me based on the name what `coercer` would do vs. `coder`.

lib/avromatic/model/custom_type.rb
```diff
@@ -0,0 +1,55 @@
...
+      def to_avro_proc
+        to_avro || value_class_method(:to_avro)
+      end
+
+      def from_avro_proc
+        from_avro || value_class_method(:from_avro)
+      end
+
+      def value_class_method(method_name)
+        value_class && value_class.respond_to?(method_name) &&
```
jturkel: We don't technically need the `value_class &&` since `nil.respond_to?(<coder method>)` should always return false.

lib/avromatic/model/passthrough_coder.rb
```diff
@@ -0,0 +1,10 @@
+module Avromatic
+  module Model
+    # This trivial coder simply returns the value provided.
+    module PassthroughCoder
```
jturkel: Nice

jturkel:
A few minor comments but looks great. I'm looking forward to seeing this cleanup our `SalsifyUuid` handling! 🚢 🎉

README.md
tjwp: Sure

lib/avromatic/model/custom_type.rb
tjwp: You're right, but I think we should keep it to be defensive. Who knows what monkey patches people will have applied to their `NilClass`!

lib/avromatic/model/custom_type.rb
tjwp: I don't like the `avro_decoder` and `avro_encoder` names since these methods aren't dealing with Avro encoded values. It's "decoding" after the Avro decode, and encoding prior to the Avro encode.

If we want something symmetric how about `deserializer` and `serializer`? 

The only other idea I have is something around conversion: `model_converter` and `avro_converter`, or `model_coercer` and `avro_coercer` (where the methods are named for the destination).

tjwp:
I have changes ready to go for the rules service once this merges.

PR #7, opened 2016-05-19
Add RuboCop and Overcommit

tjwp:
I'm surprised how many offenses this is turning up based on the config we've used with rules!

Prime: @jturkel


jturkel:
:shipit:

PR #8, opened 2016-05-24
Add simplified serialization API

tjwp:
This changes adds interfaces for simple model encoding and decoding that does not require the `AvroTurf::Messaging` API.

There are two commits here:
* first, the `serialization` module is renamed to `messaging`
* second, a new `serialization` module is implemented. Things that are shared with the `messaging` implementation were moved back to this module and new functionality implemented to handle encode/decode without a schema registry.

Prime: @jturkel

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,23 +210,23 @@ message_key = model.avro_message_key
 A model instance can be created from a key and value encoded in this manner:
 
 ```ruby
-MyTopic.deserialize(message_key, message_value)
+MyTopic.message_decode(message_key, message_value)
 ```
 
 Or just a value if only one schema is used:
 
 ```ruby
-MyValue.deserialize(message_value)
+MyValue.message_decode(message_value)
 ```
 
-#### Avromatric::Model::Decoder
+#### Avromatric::Model::MessageDecoder
 
 A stream of messages encoded from various models using the messaging approach
-can be deserialized using `Avromatic::Model::Decoder`. The decoder must be 
+can be decoded using `Avromatic::Model::MessageDecoder`. The decoder must be 
 initialized with the list of models to decode:
 
 ```ruby
-decoder = Avromatic::Model::Decoder.new(MyModel1, MyModel2)
+decoder = Avromatic::Model::MessageDecoder.new(MyModel1, MyModel2)
 
 decoder.decode(model1_messge_key, model1_message_value)
 # => instance of MyModel1
--- a/README.md
+++ b/README.md
@@ -29,22 +29,32 @@ Or install it yourself as:
 
 `Avromatic` supports the following configuration:
 
-* **schema_registry**: An `AvroTurf::SchemaRegistry` object used to store Avro schemas
-  so that they can be referenced by id. Either `schema_registry` or 
-  `registry_url` must be configured. See [Confluent Schema Registry](http://docs.confluent.io/2.0.1/schema-registry/docs/intro.html).
-* **registry_url**: URL for the schema registry. The schema registry is used to store
-  Avro schemas so that they can be referenced by id.  Either `schema_registry` or 
-  `registry_url` must be configured.
-* **schema_store**: The schema store is used to load Avro schemas from the filesystem.
+#### Model Generation
+
+* **schema_store**: A schema store is required to load Avro schemas from the filesystem.
   It should be an object that responds to `find(name, namespace = nil)` and
   returns an `Avro::Schema` object. An `AvroTurf::SchemaStore` can be used.
+  The `schema_store` is unnecessary if models are generated directly from 
+  `Avro::Schema` objects. See [Models](#models).
+  
+#### Using a Schema Registry/Messaging API
+ 
+The configuration options below are required when using a schema registry 
+(see [Confluent Schema Registry](http://docs.confluent.io/2.0.1/schema-registry/docs/intro.html)).
+and the [Messaging API](#messaging-api).
+  
+* **schema_registry**: An `AvroTurf::SchemaRegistry` object used to store Avro schemas 
+  so that they can be referenced by id. Either `schema_registry` or 
+  `registry_url` must be configured.
+* **registry_url**: URL for the schema registry. Either `schema_registry` or 
+  `registry_url` must be configured.
 * **messaging**: An `AvroTurf::Messaging` object to be shared by all generated models.
   The `build_messaging!` method may be used to create a `Messaging` instance based
   on the other configuration values.
-* logger: The logger to use for the schema registry client.
+* **logger**: The logger to use for the schema registry client.
 * [Custom Types](#custom-types)
 
-Example:
+Example using a schema registry:
 
 ```ruby
 Avromatic.configure do |config|
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ and value associated with a model.
 The attributes for the value schema used to define a model can be encoded using:
 
 ```ruby
-encoded_value = model.avro_encoded_value
+encoded_value = model.avro_raw_value
 ```
 
 In order to decode this data, a copy of the value schema is required.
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -29,16 +29,16 @@ Or install it yourself as:
 
 `Avromatic` supports the following configuration:
```
jturkel: This configuration is only required if you're using a schema registry right? If so, we should probably make some changes to the wording of this section.

README.md
```diff
@@ -29,16 +29,16 @@ Or install it yourself as:
...
   `registry_url` must be configured.
-* schema_store: The schema store is used to load Avro schemas from the filesystem.
+* **schema_store**: The schema store is used to load Avro schemas from the filesystem.
   It should be an object that responds to `find(name, namespace = nil)` and
   returns an `Avro::Schema` object. An `AvroTurf::SchemaStore` can be used.
-* messaging: An `AvroTurf::Messaging` object to be shared by all generated models.
+* **messaging**: An `AvroTurf::Messaging` object to be shared by all generated models.
   The `build_messaging!` method may be used to create a `Messaging` instance based
   on the other configuration values.
 * logger: The logger to use for the schema registry client.
```
jturkel: Not new but the `end` is missing in the block below.

README.md
```diff
@@ -136,24 +136,68 @@ inbound or outbound value is nil.
...
+
+```ruby
+encoded_key = model.avro_encoded_key
+```
+
+If attributes were encoded using the same schema(s) used to define a model, then
+the data can be decoded to create a new model instance:
+
+```ruby
+MyTopic.decode(key: encoded_key, value: encoded_value)
```
jturkel: Not new but I wonder if we should call this something like `MyEvent` instead. The term "topic" threw me a bit in this context.

README.md
jturkel: Maybe we should move it the Messaging API section?

README.md
```diff
@@ -136,24 +136,68 @@ inbound or outbound value is nil.
...
-id at the beginning of the value.
+`Avromatic` provides two different interfaces for encoding the key (optional)
+and value associated with a model.
+
+#### Manually Managed Schemas
+
+The attributes for the value schema used to define a model can be encoded using:
+
+```ruby
+encoded_value = model.avro_encoded_value
```
jturkel: Maybe we can brainstorm naming alternatives offline but I'm a bit worried that developers will confuse `avro_encoded_value` / `avro_message_value` and `decode` / `deserialize`.

jturkel:
Done with the review pass. My biggest concern is that we make it clear to developers which of the serialization/deserialization approaches they're using.

README.md
tjwp: I like keeping the configuration mostly in one place, so I'm first attempting to add more structure to this section. We can see how that works.

tjwp:
Pushed a commit with some README changes. Still need to brainstorm naming alternatives offline.

tjwp:
@jturkel Applied renaming -- ready for another pass.

README.md
```diff
@@ -29,22 +29,32 @@ Or install it yourself as:
...
   It should be an object that responds to `find(name, namespace = nil)` and
   returns an `Avro::Schema` object. An `AvroTurf::SchemaStore` can be used.
-* messaging: An `AvroTurf::Messaging` object to be shared by all generated models.
+  The `schema_store` is unnecessary if models are generated directly from 
+  `Avro::Schema` objects. See [Models](#models).
+  
+#### Using a Schema Registry/Messaging API
+ 
+The configuration options below are required when using a schema registry 
+(see [Confluent Schema Registry](http://docs.confluent.io/2.0.1/schema-registry/docs/intro.html)).
```
jturkel: extra period

README.md
```diff
@@ -29,22 +29,32 @@ Or install it yourself as:
...
+* **schema_registry**: An `AvroTurf::SchemaRegistry` object used to store Avro schemas 
+  so that they can be referenced by id. Either `schema_registry` or 
+  `registry_url` must be configured.
+* **registry_url**: URL for the schema registry. Either `schema_registry` or 
+  `registry_url` must be configured.
+* **messaging**: An `AvroTurf::Messaging` object to be shared by all generated models.
   The `build_messaging!` method may be used to create a `Messaging` instance based
   on the other configuration values.
-* logger: The logger to use for the schema registry client.
+* **logger**: The logger to use for the schema registry client.
```
jturkel: Should `logger` be a common configuration option across the two encoding schemes?

README.md
```diff
@@ -29,22 +29,32 @@ Or install it yourself as:
...
 * [Custom Types](#custom-types)
 
-Example:
+Example using a schema registry:
 
 ```ruby
```
jturkel: This block is missing an `end`

README.md
```diff
@@ -136,59 +146,103 @@ inbound or outbound value is nil.
...
+
+```ruby
+encoded_key = model.avro_raw_key
+```
+
+If attributes were encoded using the same schema(s) used to define a model, then
+the data can be decoded to create a new model instance:
+
+```ruby
+MyTopic.raw_decode(key: encoded_key, value: encoded_value)
```
jturkel: Should we use a name like `MyModel` that is less tied to Kafka (and the assumption that topics only have a single type associated with them)?

lib/avromatic/model/serialization.rb
```diff
@@ -1,30 +1,24 @@
...
 require 'avromatic/model/passthrough_serializer'
 
 module Avromatic
   module Model
 
-    # This concern adds support for serialization to a model
-    # generated from Avro schema(s).
+    # This module provides serialization support for encoding directly to Avro
+    # without dependency on a schema registry.
     module Serialization
```
jturkel: Should we rename this module to something like `Raw` or `RawSerialization`? If we go with the latter we should probably also rename `Messaging` to `MessagingSerialization` for consistency.

lib/avromatic/model/messaging.rb
```diff
@@ -0,0 +1,57 @@
...
+      include Encode
+
+      # This module provides methods to message_decode an Avro-encoded value and
+      # an optional Avro-encoded key as a new model instance.
+      module Decode
+
+        # If two arguments are specified then the first is interpreted as the
+        # message key and the second is the message value. If there is only one
+        # arg then it is used as the message value.
+        def message_decode(*args)
```
jturkel: Should this be prefixed with `avro_` to minimize the likelihood of a collision with domain specific names?

lib/avromatic/model/messaging.rb
```diff
@@ -0,0 +1,57 @@
...
+          value_attributes = messaging.decode(message_value, schema_name: avro_schema.fullname)
+
+          new(value_attributes.merge!(key_attributes || {}))
+        end
+      end
+
+      module ClassMethods
+        # The messaging object acts as an intermediary talking to the schema
+        # registry and using returned/specified schemas to decode/encode.
+        delegate :messaging, to: :Avromatic
```
jturkel: Same question about using the `avro_` prefix.

lib/avromatic/model/serialization.rb
```diff
@@ -50,38 +42,69 @@ def avro_hash(fields)
                                end
           end
         end
+
+        def encode(data, key_or_value = :value)
```
jturkel: This module seems to have quite a few methods that will be added to generated models. Should they either be prefixed with `avro_` or put in a separate helper module that is not included in generated models?

jturkel:
A few additional questions. The biggest area of concern is polluting generated models with lots of avromatic generated methods but we can always tackle that in a separate PR. :shipit:

README.md
tjwp: It's only used for the schema registry, which is only used with the messaging approach, so I think it should stay here until its needed more generally.

lib/avromatic/model/messaging.rb
tjwp: Or should it be _avromatic__? I'm being facetious. I'm okay renaming this.

lib/avromatic/model/serialization.rb
tjwp: Added some additional prefixing.

PR #10, opened 2016-06-10
Fix serialization of array or map of nested models

tjwp:
New tests verify that models of these forms can be initialized using a hash, and round-trip correctly through serialization.

Prime: @jturkel


jturkel:
:shipit:

PR #12, opened 2016-06-14
Remove dependency on transform_values from ActiveSupport v4.2

### LINKED ISSUE #11: Avromatic doesn't work with Rails 4.1
`lib/avromatic/model/raw_serialization.rb` is requiring `active_support/core_ext/hash/transform_values` which was added in Rails 4.2.

tjwp:
Fixes #11

Prime: @jturkel


jturkel:
:shipit:

PR #13, opened 2016-06-14
Add test configurations for different ActiveModel/ActiveSupport versions

tjwp:
Start by adding configurations for ruby and Rails versions that Salsify if using.

Prime: @jturkel


jturkel:
:shipit:

Issue #16, opened 2016-07-21
Are Avro Union fields supported?


cyclops23:
Trying to generate a model for this example schema that uses a Union of Foo and Bar...
```
require 'avromatic'

json = '[{
    "type": "record",
    "name": "Foo",
    "fields": [
      {"name": "fooMessage", "type": "string"}
    ]
  },

  {
    "type": "record",
    "name": "Bar",
    "fields": [
      {"name": "barMessage", "type": "string"}
    ]
  },

  {
    "type": "record",
    "name": "Root",
    "fields": [
      {"name": "header", "type": "string"},
      {"name": "message", "type": ["Foo", "Bar"]}
    ]
  }
]
'
Avromatic::Model.model(schema: Avro::Schema.parse(json))
```
...I encounter an exception:
```
/Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/attributes.rb:54:in `define_avro_attributes': undefined method `fields' for #<Avro::S
chema::UnionSchema:0x007f9b789eb040> (NoMethodError)
        from /Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/attributes.rb:27:in `add_avro_fields'
        from /Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/builder.rb:57:in `block (2 levels) in define_included_method'
        from /Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/builder.rb:22:in `include'
        from /Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/builder.rb:22:in `block in model'
        from /Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/builder.rb:21:in `initialize'
        from /Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/builder.rb:21:in `new'
        from /Users/ascassidy/.rvm/gems/ruby-2.3.0/gems/avromatic-0.7.0/lib/avromatic/model/builder.rb:21:in `model'
[...]

PR #18, opened 2016-08-18
Raise an error for non-record Avro types

tjwp:
This change adds test cases and a more descriptive error when the top-level type used to define a model is not an Avro record.

This provides better feedback for the schema in Issue #16.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -51,6 +51,10 @@ module Avromatic
         end
 
         def define_avro_attributes(schema)
+          if schema.type_sym != :record
+            raise "Unsupported schema type '#{schema.type_sym}', only 'record' schemas are supported."
+          end
+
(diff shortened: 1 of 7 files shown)


lib/avromatic/model/attributes.rb
```diff
@@ -51,6 +51,10 @@ def schema_fields_differ?(name)
         end
 
         def define_avro_attributes(schema)
+          if schema.type_sym != :record
```
jturkel: Minor but this could be an `unless` if you want to avoid the `!=`.

jturkel:
:shipit:

lib/avromatic/model/attributes.rb
tjwp: I wrote it that way initially then decided to avoid the `unless` instead. Lesser of two evils ...?

Issue #17, opened 2016-08-16
related with virtus: "attributes" is not allowed as an attribute name 


swistaczek:
Some AVRO data model managed by other team is using field named ""attributes" which collided with virtus restricted [attributes name](https://github.com/solnic/virtus/blob/master/spec/unit/virtus/attribute_spec.rb#L40).

```ruby
ArgumentError: "attributes" is not allowed as an attribute name
       	from /Users/ernest.bursa/.rvm/gems/ruby-2.3.1@icet-self-service/gems/virtus-1.0.5/lib/virt
```

tjwp:
Some prototyping on a solution for this was done here: https://github.com/salsify/avromatic/pull/19

It's likely that in a final form the configuration for this will change. Ideally, the renaming or mapping would be driven off fields names within named record types and configured globally, so that configuration does not have to be repeated, and with the possibility of overrides for a specific model.

PR #19, opened 2016-08-22
WIP: Support field renaming

tjwp:
I did some work on allowing schema fields to be renamed in the generated model. This would to be to address Issue #17, but may also help if a schema field name conflicts with another method on an Avromatic model.

If this goes forward, I think that "aliases" should be changed to something like "renames" to avoid confusion with aliases in Avro schema names, and to make it clear the original field name is not available on the model.

It's also unclear what is the most natural way to configure this. In this code, the configuration is based on the field names within a schema, including nested fields. Would it be better to configure renaming scoped by the record type name instead of the field name within a record? 

There still some work to do on validating the renames/aliases.

These changes also raise the question of whether there are too many methods defined on model instances by Avromatic. For example, there is a set of methods delegated to the class. Should these instead be called directly on the class to avoid field conflicts, or should a prefix be used to reduce the chance of a name collision. This should probably be handled separately.

I think the changes to `Avromatic::Model::Attributes` here also suggest that this module should be refactored. Instead of passing around information about the field/attribute currently being processed we could introduce an `AttributeBuilder` class that holds the context and responds to methods providing what should be generated for that attribute.

Prime: @jturkel


README.md
```diff
@@ -148,6 +148,30 @@ inbound or outbound value is nil.
...
+defining the model.
+
+For example, `attributes` is a reserved name for Virtus (used internally by
+Avromatic). If a schema contains an attributes field, then it can be aliased
+to a different name on the model:
+
+```ruby
+Avromatic::Model.model(
+  schema_name: 'reserved_names',
+  aliases: { attributes: :my_attributes }
```
jturkel: I agree `aliases` is a bit misleading. `renames` or `mappings` work for me.

lib/avromatic/model/attributes.rb
```diff
@@ -67,6 +66,40 @@ def define_avro_attributes(schema)
...
+        def validate_schema!(schema)
+          if schema.type_sym != :record
+            raise "Unsupported schema type '#{schema.type_sym}', only 'record' schemas are supported."
+          end
+
+          invalid_fields = schema.fields.select do |field|
+            instance_methods.include?(field.name.to_sym) && no_valid_alias?(field.name)
+          end
+          if invalid_fields.any?
+            raise "Disallowed field names: #{invalid_fields.map(&:name).map(&:inspect).join(', ')}.\n"\
```
jturkel: Nice! This will avoid some really hard to find bugs.

lib/avromatic/model/configurable.rb
```diff
@@ -8,14 +8,19 @@ module Configurable
 
       module ClassMethods
         attr_accessor :config
-        delegate :avro_schema, :value_avro_schema, :key_avro_schema, to: :config
+        delegate :avro_schema, :value_avro_schema, :key_avro_schema,
+                 :aliases, :inverse_aliases, to: :config
+
+        def inherited(subclass)
+          subclass.config = config
```
jturkel: Does this need to be an `extended` hook?

tjwp: I think it is `inherited` because the generated model class owns the configuration. If that model is subclassed then we want the subclass to refer back to the same configuration. 

I guess by definition that's a class variable, so we could switch to that for the config on a generated model class.

tjwp:
Putting this on hold for now.

PR #20, opened 2016-08-26
Add appraisal

tjwp:
This change is being made in advance of adding logical type support to avromatic.

Previously a matrix of builds was being used with Travis to test Rails 4.1 and 4.2. This switches to using `appraisal` on Travis. Rails 4.1 & 4.2 are tested with the latest official `avro` gem and Rails 4.2 is tested with `avro-salsify-fork`.

Prime: @jturkel


jturkel:
:shipit:

PR #21, opened 2016-08-29
Add support for logical types

tjwp:
The changes here to support the three logical types that make sense with builtin Ruby types were relatively straightforward. But the actual use of these logical types found some issues with the proposed changes to the official avro gem: https://github.com/apache/avro/pull/116. This required a new release of our fork.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/logical_types.rb
+++ b/lib/avromatic/model/logical_types.rb
@@ -1,0 +1,16 @@
+module Avromatic
+  module Model
+    module LogicalTypes
+
+      LOGICAL_TYPE_MAP = {
+        'date' => Date,
+        'timestamp-micros' => Time,
+        'timestamp-millis' => Time
+      }.freeze
+
+      def self.value_class(logical_type)
+        LOGICAL_TYPE_MAP.fetch(logical_type, nil)
+      end
+    end
+  end
+end
--- a/lib/avromatic/model/logical_types.rb
+++ b/lib/avromatic/model/logical_types.rb
@@ -9,7 +9,7 @@ module Avromatic
       }.freeze
 
       def self.value_class(logical_type)
-        LOGICAL_TYPE_MAP.fetch(logical_type, nil)
+        LOGICAL_TYPE_MAP[logical_type]
       end
     end
   end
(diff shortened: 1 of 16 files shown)


lib/avromatic/model/logical_types.rb
```diff
@@ -0,0 +1,16 @@
...
+    module LogicalTypes
+
+      LOGICAL_TYPE_MAP = {
+        'date' => Date,
+        'timestamp-micros' => Time,
+        'timestamp-millis' => Time
+      }.freeze
+
+      def self.value_class(logical_type)
+        LOGICAL_TYPE_MAP.fetch(logical_type, nil)
```
jturkel: This could probably just be `LOGICAL_TYPE_MAP[logical_type]` since you're falling back to `nil` anyway.

spec/spec_helper.rb
```diff
@@ -1,11 +1,18 @@
 $LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
 require 'simplecov'
 
-SimpleCov.start
+SimpleCov.start do
+  add_filter 'spec'
+  minimum_coverage 99
```
jturkel: Cheater ;)

spec/support/contexts/logical_types_serialization.rb
```diff
@@ -0,0 +1,45 @@
...
+          expect(decoded).to eq(instance)
+        end
+      end
+    end
+
+    without_logical_types do
+      context "unsupported" do
+        let(:values) do
+          {
+            date: (Date.today - Date.new(1970, 1, 1)).to_i,
```
jturkel: Might be worth a `let` that describes the intent of `Date.new(1970, 1, 1)`

jturkel:
A few minor comments. :shipit:

lib/avromatic/model/logical_types.rb
tjwp: For some reason when I wrote this this felt like it captured more explicitly that its okay for a logical type to not be present, and we're expecting to handle nil.

PR #23, opened 2016-08-31
Add full union support

tjwp:
This change addresses Issue #23.

My plan is to do a pre-release with these changes as I expect we'll discover additional edge cases through real use.

One unaddressed problem with union support is the use of a custom type within a real union. I don't believe we have an immediate use case for that, so I intend to defer that.

These changes also suggest that `Avromatic::Model::Attributes` should be refactored so that less state must be passed around. I think that should also be handled in a separate PR.

Prime: @jturkel


lib/avromatic/model/attribute/union.rb
```diff
@@ -0,0 +1,46 @@
+require 'avromatic/model/attribute_type/union'
+
+module Avromatic
+  module Model
+    module Attribute
+      class Union < Virtus::Attribute
```
jturkel: Probably worth a few class level comments on how this integrates into the Virtus framework.

lib/avromatic/model/attribute/union.rb
```diff
@@ -0,0 +1,46 @@
...
+          primitive.types.each do |type|
+            member_attributes << Virtus::Attribute.build(type)
+          end
+        end
+
+        def coerce(input)
+          return input if value_coerced?(input)
+
+          result = nil
+          member_attributes.find do |union_attribute|
```
jturkel: Do we want `result = member_attributes.find` to properly handle the case of an invalid attribute (since we assign to `result` before the validation check on line 23)?

lib/avromatic/model/attribute_type/union.rb
```diff
@@ -0,0 +1,25 @@
+module Avromatic
+  module Model
+    module AttributeType
+      class Union
```
jturkel: Probably worth a few class level comments on how this integrates into the Virtus framework.

lib/avromatic/model/attribute_type/union.rb
```diff
@@ -0,0 +1,25 @@
...
+          attr_writer :types
+        end
+
+        # Factory method to define Union types with the specified list of
+        # types (classes).
+        def self.[](*types)
+          Class.new(self).tap do |klass|
+            klass.types = types
+            # See https://github.com/solnic/virtus/issues/284#issuecomment-56405137
+            Axiom::Types::Object.new { primitive(klass) }
```
jturkel: Does this add the type to some kind of type registry?

lib/avromatic/model/attributes.rb
```diff
@@ -127,24 +128,33 @@ def avro_field_class(field_type)
...
-            end
+            Avromatic::Model.model(schema: field_type)
           else
             raise "Unsupported type #{field_type}"
           end
         end
 
         def union_field_class(field_type)
-          avro_field_class(Avromatic::Model::Attributes.first_union_schema(field_type))
+          field_classes = field_type.schemas.select { |schema| schema.type_sym != :null }
```
jturkel: Could probably do a `reject` rather than a `select` here.

lib/avromatic/model/attributes.rb
```diff
@@ -127,24 +128,33 @@ def avro_field_class(field_type)
...
 
-        def avro_field_options(field)
+        def avro_field_options(field, field_class)
           options = {}
 
-          custom_type = Avromatic.type_registry.fetch(field)
-          coercer = custom_type.deserializer
-          options[:coercer] = coercer if coercer
+          if field_class.is_a?(Class) && field_class < Avromatic::Model::AttributeType::Union
+            # TODO: pass coercer for member attributes in union
```
jturkel: Should we raise an error if there's a member attribute with a custom coercer until we add proper support for this?

lib/avromatic/model/attributes.rb
```diff
@@ -154,8 +164,9 @@ def avro_field_options(field)
           options
         end
 
-        def add_serializer(field)
-          custom_type = Avromatic.type_registry.fetch(field)
+        def add_serializer(field, field_class)
+          # TODO: handle serializer for custom type within union
```
jturkel: Similar question as earlier about raising an error if someone tries to do something we don't yet support

jturkel:
A few comments/question but overall looks awesome. :shipit:

lib/avromatic/model/attribute/union.rb
tjwp: I don't think we ever want the result of `find` since that would be one of the `member_attributes`. But I'll introduce a local variable to only assign to `result` if condition is met.

lib/avromatic/model/attribute_type/union.rb
tjwp: Yes, best I can tell this is caching a mapping between an Axiom type and the class. This prevents Virtus from falling back to `BasicObject` as the primitive class for these attributes.

lib/avromatic/model/attributes.rb
tjwp: Done

lib/avromatic/model/attributes.rb
tjwp: Should never hit an error here, since the error for adding a coercer comes first, but done.

PR #24, opened 2016-09-02
Fix the handling of nested, complex types

tjwp:
There are two basic issues addressed here:
- the serialization of arrays/maps of unions
- the coercion of records within other complex types

The latter issue seems to be a problem with Virtus. An array of generated models is coerced correctly, but an array of arrays of generated models is not coerced (for example). 

To workaround this issue a similar solution to that used for union support was implemented. Each generated model is registered with axiom and there is a new `Virtus::Attribute`subclass that ensures generated models are coerced by checking for a model instance or creating and validating a new one.

The plan once again is to create a pre-release with these changes for further testing.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -128,7 +128,9 @@ module Avromatic
           when :record
             # TODO: This should add the generated model to a module.
             # A hash of generated models should be kept by name for reuse.
-            Avromatic::Model.model(schema: field_type)
+            Avromatic::Model.model(schema: field_type).tap do |record_class|
+              Axiom::Types::Object.new { primitive(record_class) }
+            end
           else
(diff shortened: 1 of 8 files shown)


lib/avromatic/model/attributes.rb
```diff
@@ -128,7 +128,9 @@ def avro_field_class(field_type)
           when :record
             # TODO: This should add the generated model to a module.
             # A hash of generated models should be kept by name for reuse.
-            Avromatic::Model.model(schema: field_type)
+            Avromatic::Model.model(schema: field_type).tap do |record_class|
+              Axiom::Types::Object.new { primitive(record_class) }
```
jturkel: This could probably use a comment

jturkel:
:shipit:

PR #25, opened 2016-09-10
WIP: Reuse generated models

tjwp:
This change adds support for reusing and directly referencing embedded models.

The `Avromatic` module can be configured with a *module* that is used to collect references to embedded models. These models are stored under a module namespace that is generated based on their schema namespace. There is also a *strip_namespace_prefix* option that can be configured to remove a prefix from the schema namespace before generating modules for namespacing.

If a module is not configured, then a new anonymous module is used.

The *module* and *strip_namespace_prefix* option can both be overridden when generating a model, otherwise generated models inherit from the `Avromatic` module.

Additionally, embedded models are available on a model via a `models` hash that is keyed by the field name.

Prime: @jturkel -- looking for high-level feedback on the form factor, API, and general approach.

Needs additional documentation and specs.

### DIFF
--- a/lib/avromatic/model/configuration.rb
+++ b/lib/avromatic/model/configuration.rb
@@ -4,7 +4,7 @@ module Avromatic
     # This class holds configuration for a model built from Avro schema(s).
     class Configuration
 
-      attr_reader :avro_schema, :key_avro_schema
+      attr_reader :avro_schema, :key_avro_schema, :module, :strip_namespace_prefix
       delegate :schema_store, to: Avromatic
 
       # Either schema(_name) or value_schema(_name), but not both, must be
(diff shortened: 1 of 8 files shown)


lib/avromatic/model/configuration.rb
```diff
@@ -17,10 +17,14 @@ class Configuration
...
       # @option options [Avro::Schema] :key_schema
       # @option options [String, Symbol] :key_schema_name
+      # @option options [Module] :module
+      # @option options [String, Regexp] :strip_namespace_prefix
       def initialize(**options)
         @avro_schema = find_avro_schema(**options)
         raise ArgumentError.new('value_schema(_name) or schema(_name) must be specified') unless avro_schema
         @key_avro_schema = find_schema_by_option(:key_schema, **options)
+        @module = options[:module]
+        @strip_namespace_prefix = options[:@strip_namespace_prefix]
```
tjwp: The key here is broken.

tjwp:
Switching to a hash-based implementation.

PR #26, opened 2016-09-14
Reuse nested models

tjwp:
Nested models are stored in a `ModelRegistry` (`Avromatic.nested_models`). This registry is inherited by generated models, but can also be overridden. Models in the registry can be retrieved by schema full name, optionally with a namespace prefix removed.

Also, added README info on experimental union support.

Prime: @jturkel

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,11 +119,15 @@ Avromatic contains experimental support for unions containing more than one
 non-null member type. This feature is experimental because Virtus attributes
 may attempt to coerce between types too aggressively.
 
+For now, if a union contains [nested models](#nested-models) then it is
+recommended that you assign model instances.
+
+Some combination of the ordering of member types in the union and relying on
+model validation may be required so that the correct member is selected,
+especially when deserializing from Avro.
+
 In the future, the type coercion used in the gem will be replaced to better
 support the union use case.
-
-For now, if a union contains nested models then it is recommended that you
-assign instances
 
 #### Nested Models
 
@@ -145,6 +149,8 @@ The `ModelRegistry` can be customized to remove a namespace prefix:
 Avromatic.nested_models =
   Avromatic::ModelRegistry.new(remove_namespace_prefix: 'com.my_company'
 ```
+
+The `:remove_namespace_prefix` value can be a string or a regexp.
 
 By default, top-level generated models reuse `Avromatic.nested_models`. This
 allows nested models to be shared across different generated models.
@@ -169,6 +175,9 @@ class UsefulSubrecord
 end
 Avromatic.nested_models.register(UsefulSubrecord)
 ```
+
+With Rails, it may be necessary to perform this explicit registration in an
+initializer so that lazy class loading works correctly in development.
 
 #### Custom Types
 
--- a/README.md
+++ b/README.md
@@ -37,7 +37,11 @@ Or install it yourself as:
   returns an `Avro::Schema` object. An `AvroTurf::SchemaStore` can be used.
   The `schema_store` is unnecessary if models are generated directly from 
   `Avro::Schema` objects. See [Models](#models).
-  
+* **nested_models**: An optional [ModelRegistry](https://github.com/salsify/avromatic/blob/master/lib/avromatic/model_registry.rb)
+  that is used to store, by full schema name, the generated models that are
+  embedded within top-level models. By default a new `Avromatic::ModelRegistry`
+  is created.
+
 #### Using a Schema Registry/Messaging API
  
 The configuration options below are required when using a schema registry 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -109,6 +113,63 @@ constant:
 MyModel = Avromatic::Model.model(schema_name :my_model)
 ```
 
+#### Experimental: Union Support
+
+Avromatic contains experimental support for unions containing more than one
+non-null member type. This feature is experimental because Virtus attributes
+may attempt to coerce between types too aggressively.
```
jturkel: Are there more specifics we can add here so users can tell if the feature will work for them?

README.md
```diff
@@ -109,6 +113,63 @@ constant:
...
+
+In the future, the type coercion used in the gem will be replaced to better
+support the union use case.
+
+For now, if a union contains nested models then it is recommended that you
+assign instances
```
jturkel: "model instances" rather than "instances"? Also missing a period.

README.md
```diff
@@ -109,6 +113,63 @@ constant:
...
+
+```ruby
+class UsefulSubrecord
+  include Avromatic::Model.build(schema_name: 'useful_subrecord')
+
+  def do_something_custom
+    ...
+  end
+end
+Avromatic.nested_models.register(UsefulSubrecord)
```
jturkel: Is it worth some comments about doing this registration in an initializer so lazy class loading will work?

lib/avromatic/model/configuration.rb
```diff
@@ -17,10 +17,12 @@ class Configuration
       # @option options [String, Symbol] :value_schema_name
       # @option options [Avro::Schema] :key_schema
       # @option options [String, Symbol] :key_schema_name
+      # @option options [Hash] :nested_models
```
jturkel: Should this be a `Avromatic::ModelRegistry` rather than a `Hash`?

lib/avromatic/model/nested_models.rb
```diff
@@ -0,0 +1,26 @@
...
+      module ClassMethods
+        def build_nested_model(schema)
+          fullname = schema.fullname
+
+          if nested_models.model?(fullname)
+            nested_models[fullname]
+          else
+            nested_model = Avromatic::Model.model(schema: schema,
+                                                  nested_models: nested_models)
+            Axiom::Types::Object.new { primitive(nested_model) }
```
jturkel: This isn't new but can we add a comment on why we need to do this (since the side-effects are non-obvious)?

lib/avromatic/model/value_object.rb
```diff
@@ -1,5 +1,6 @@
 module Avromatic
   module Model
+    extend ActiveSupport::Concern
```
jturkel: Do we need this?

lib/avromatic/model_registry.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    def register(model)
+      raise 'models with a key schema are not supported' if model.key_avro_schema
+      name = model.avro_schema.fullname
+      name = remove_prefix(name) if @prefix
+      @hash[name] = model
+    end
+
+    def model?(fullname)
```
jturkel: Calling this `registered?` would make the naming more consistent with `register` and might make code at calling sites a bit more readable.

lib/avromatic/model/configuration.rb
tjwp: Good catch.

lib/avromatic/model_registry.rb
```diff
@@ -0,0 +1,43 @@
...
+      @hash.key?(fullname)
+    end
+
+    private
+
+    def remove_prefix(name)
+      value =
+        case @prefix
+        when String
+          name[@prefix.length..-1] if name.start_with?(@prefix)
```
jturkel: nit: Could do `name.from(@prefix.length)` if we're already relying on ActiveSupport.

lib/avromatic/model/value_object.rb
tjwp: Nope - added for something I pulled out.

lib/avromatic/model_registry.rb
```diff
@@ -0,0 +1,43 @@
...
+        when Regexp
+          name.sub(@prefix, '')
+        else
+          raise "unsupported `remove_namespace_prefix` value: #{@prefix}"
+        end
+
+      value.start_with?('.') ? value[1..-1] : value
```
jturkel: Same nit about using `from`

README.md
```diff
@@ -109,6 +113,63 @@ constant:
...
+```ruby
+Avromatic.nested_models['com.my_company.test.example']
+#=> <model class>
+```
+
+The `ModelRegistry` can be customized to remove a namespace prefix:
+
+```ruby
+Avromatic.nested_models =
+  Avromatic::ModelRegistry.new(remove_namespace_prefix: 'com.my_company'
```
jturkel: Should we doc that this can be a regex too?

spec/avromatic/model/builder_nested_models_spec.rb
```diff
@@ -0,0 +1,65 @@
+describe Avromatic::Model::Builder, 'nested_models' do
```
jturkel: Do we use the `nested_models` tag for anything?

spec/avromatic/model_registry_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+  describe "#model?" do
+    let(:instance) { described_class.new }
+
+    context "for a model that has not been registered" do
+      it "returns false" do
+        expect(instance.model?('test.nested_record')).to eql(false)
+      end
+    end
+
+    context "for  model that has been registered" do
```
jturkel: extra space

spec/avromatic/model/builder_nested_models_spec.rb
tjwp: Technically I think it is "additional description" here, not a tag. 

This was an attempt to break up the specs a little from their mostly monolithic form. 

The only way this additional description is used is in the test output (subject to formatter) and aligning the the spec file path, as enforced by rubocop-rspec.

jturkel:
All minor feedback. Looks great. :shipit:

lib/avromatic/model_registry.rb
tjwp: Done

lib/avromatic/model_registry.rb
tjwp: in for a penny, in for a pound

README.md
tjwp: Done

tjwp: Added

README.md
tjwp: Done

README.md
tjwp: Included some additional tips.

PR #27, opened 2016-09-15
WIP: Nested model methods spike

tjwp:
Experiment with adding methods to return nested models.

Method missing is used initially to lookup the model. If a model is found, then a method is defined on the current model.

To make the mapping between schema full names and methods more predictable, periods in full names are replaced by two underscores:
- For the schema 'admin.user': `admin__user_model`
- For the schema 'admin_user': `admin_user_model` (hopefully an organization doesn't use both!)

(This means that double underscores and trailing underscores are not supported in schema names.)

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/nested_models.rb
+++ b/lib/avromatic/model/nested_models.rb
@@ -23,6 +23,32 @@ module Avromatic
             nested_models.register(nested_model)
           end
         end
+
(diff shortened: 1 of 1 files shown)


tjwp:
Putting this on hold for now.

PR #31, opened 2016-09-23
Allow empty required array and map fields

tjwp:
Previously a presence check was used for all required fields. This check uses `blank?` which has the effect of marking empty arrays and maps as invalid.

As part of this change, the validation specs were moved into a separate file.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -85,7 +85,7 @@ module Avromatic
         def add_required_validation(field)
           if required?(field) && field.default == :no_default
             case field.type.type_sym
-            when :array, :map
+            when :array, :map, :boolean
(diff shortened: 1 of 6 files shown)


lib/avromatic/model/attributes.rb
```diff
@@ -84,7 +84,12 @@ def add_validation(field)
 
         def add_required_validation(field)
           if required?(field) && field.default == :no_default
-            validates(field.name, presence: true)
+            case field.type.type_sym
+            when :array, :map
+              validates(field.name, exclusion: { in: [nil], message: "can't be nil" })
```
jturkel: Do we want to use this validation approach for booleans too?

tjwp: Looks like the behavior was already correct here, but I switched booleans to the same nil check.

PR #34, opened 2016-09-26
Only allow nested models to be registered once

tjwp:
This change only allows a model's full name to be registered once for the same registry. Any attempt to register the same name again raises an error.

Prime: @jturkel


jturkel:
:shipit:

PR #36, opened 2016-09-26
Nested validation of records embedded in complex types

tjwp:
This change completes validation for all of the places where nested models can appear. 

Instead of attempting to predict (based on the schema) where nested, complex types will occur, this implementation checks any complex types recursively.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/validation.rb
+++ b/lib/avromatic/model/validation.rb
@@ -1,0 +1,48 @@
+module Avromatic
+  module Model
+    module Validation
+      extend ActiveSupport::Concern
+
+      module ClassMethods
+
+        # Returns an array of messages
+        def validate_nested_value(value)
+          case value
+          when Avromatic::Model::Attributes
+            validate_record_value(value)
+          when Array
+            value.map.with_index do |element, index|
+              validate_nested_value(element).map do |message|
+                "[#{index}]#{message}"
+              end
+            end.flatten
+          when Hash
+            value.map do |key, map_value|
+              validate_nested_value(map_value).map do |message|
+                "['#{key}']#{message}"
+              end
+            end.flatten
+          end || []
+        end
+
+        private
+
+        def validate_complex(field_name)
+          validate do |instance|
+            value = instance.send(field_name)
+            messages = self.class.validate_nested_value(value)
+            messages.each { |message| instance.errors.add(field_name.to_sym, message) }
+          end
+        end
+
+        def validate_record_value(record)
+          if record && record.invalid?
+            record.errors.map do |key, message|
+              ".#{key} #{message}".gsub(' .', '.')
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/avromatic/model/validation.rb
+++ b/lib/avromatic/model/validation.rb
@@ -11,18 +11,22 @@ module Avromatic
           when Avromatic::Model::Attributes
             validate_record_value(value)
           when Array
-            value.map.with_index do |element, index|
+            value.flat_map.with_index do |element, index|
               validate_nested_value(element).map do |message|
                 "[#{index}]#{message}"
               end
-            end.flatten
+            end
           when Hash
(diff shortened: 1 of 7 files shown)


lib/avromatic/model/validation.rb
```diff
@@ -0,0 +1,48 @@
...
+        def validate_nested_value(value)
+          case value
+          when Avromatic::Model::Attributes
+            validate_record_value(value)
+          when Array
+            value.map.with_index do |element, index|
+              validate_nested_value(element).map do |message|
+                "[#{index}]#{message}"
+              end
+            end.flatten
```
jturkel: Does `flat_map.with_index` work?

lib/avromatic/model/validation.rb
```diff
@@ -0,0 +1,48 @@
...
+          when Hash
+            value.map do |key, map_value|
```
jturkel: Is it worth a comment the hash keys are always strings in Avro so they require no validation (although I'm not sure where validating type coercion fits in for hash keys)?

lib/avromatic/model/validation.rb
```diff
@@ -0,0 +1,48 @@
...
+              validate_nested_value(map_value).map do |message|
+                "['#{key}']#{message}"
+              end
+            end.flatten
+          end || []
```
jturkel: Might be a little clearer to add an `else` to the `case` rather than the `|| []`.

lib/avromatic/model/validation.rb
```diff
@@ -0,0 +1,48 @@
...
+            value = instance.send(field_name)
+            messages = self.class.validate_nested_value(value)
+            messages.each { |message| instance.errors.add(field_name.to_sym, message) }
+          end
+        end
+
+        def validate_record_value(record)
+          if record && record.invalid?
+            record.errors.map do |key, message|
+              ".#{key} #{message}".gsub(' .', '.')
```
jturkel: Why do we need this `gsub`?

lib/avromatic/model/validation.rb
```diff
@@ -0,0 +1,48 @@
...
+
+        # Returns an array of messages
+        def validate_nested_value(value)
+          case value
+          when Avromatic::Model::Attributes
+            validate_record_value(value)
+          when Array
+            value.map.with_index do |element, index|
+              validate_nested_value(element).map do |message|
+                "[#{index}]#{message}"
```
jturkel: What do the full error messages look like?

lib/avromatic/model/validation.rb
tjwp: Coercion for hash keys should happen due to the attribute type used with Virtus: `Hash[String => ???]`.

lib/avromatic/model/validation.rb
tjwp: Also requires an else in `validate_record_value` as previously that could return nil but I've made the change.

lib/avromatic/model/validation.rb
tjwp: Without the `gsub`, errors for nested records could be returned as: ".foo .bar .baz can't be blank". The `gsub` compacts that to ".foo.bar.baz".

lib/avromatic/model/validation.rb
tjwp: Full error messages roughly have the form: "#{field_name.humanize} #{message}", so a humanized top-level field name will be prepended to whatever is returned here. 

Some examples:
```
# top-level array "my_widgets", where widgets are records with a required name and price:
My widgets [1].name can't be blank

# Top-level hash, "all_widgets", by name with a missing price:
All widgets ['sprocket'].price can't be blank

# For a top-level "this_widget", a parent without a price:
This widget parent.price can't be blank

# For a top-level union "this_or_that":
This or that price can't be blank
```

tjwp:
Responded to questions

jturkel:
Still looks good. :shipit:

PR #37, opened 2016-09-27
Coerce timestamps to the supported precision

tjwp:
With this change, instances match exactly after a round-trip through serialization and deserialization.

Previously a timestamp stored in a model could have greater precision than a value returned after deserialization.

This removes any need to truncate timestamps before assigning to an Avromatic model.

Prime: @jturkel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@
 - Add support for recursive models.
 - Allow required array and map fields to be empty. Only nil values for required
   array and map fields are now considered invalid.
-- Validate nested models. This does not apply to nested models embedded
-  within other complex types (array, map, and union).
+- Truncate values for timestamps to the precision supported by the logical type.
 
 ## v0.8.0
 - Add support for logical types. Currently this requires using the
--- a/lib/avromatic/model/attribute/abstract_timestamp.rb
+++ b/lib/avromatic/model/attribute/abstract_timestamp.rb
@@ -1,0 +1,26 @@
+module Avromatic
+  module Model
+    module Attribute
+
+      # This subclass of Virtus::Attribute is used to truncate timestamp values
+      # to the supported precision.
+      class AbstractTimestamp < Virtus::Attribute
+        def coerce(value)
+          return value if value.nil? || value_coerced?(value)
+
+          value.is_a?(Time) ? coerce_time(value) : value
+        end
+
+        def value_coerced?(_value)
+          raise 'subclass must implement `value_coerced?`'
+        end
+
+        private
+
+        def coerce_time(_value)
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -7,8 +7,7 @@
 - Add support for recursive models.
 - Allow required array and map fields to be empty. Only nil values for required
   array and map fields are now considered invalid.
-- Validate nested models. This does not apply to nested models embedded
```
jturkel: Merge error?

lib/avromatic/model/attribute/abstract_timestamp.rb
```diff
@@ -0,0 +1,26 @@
...
+      # to the supported precision.
+      class AbstractTimestamp < Virtus::Attribute
+        def coerce(value)
+          return value if value.nil? || value_coerced?(value)
+
+          value.is_a?(Time) ? coerce_time(value) : value
+        end
+
+        def value_coerced?(_value)
+          raise 'subclass must implement `value_coerced?`'
```
jturkel: We would get some more useful context if the message was `"#{__method__} must be overridden by #{self.class.name}"`

lib/avromatic/model/attribute/timestamp_micros.rb
```diff
@@ -0,0 +1,23 @@
...
+      class TimestampMicros < AbstractTimestamp
+
+        def value_coerced?(value)
+          value.is_a?(Time) && value.nsec % 1000 == 0
+        end
+
+        private
+
+        def coerce_time(value)
+          Time.at(value.to_i, value.usec)
```
jturkel: How does this work with timezones?

lib/avromatic/model/attribute/timestamp_millis.rb
```diff
@@ -0,0 +1,23 @@
...
+      class TimestampMillis < AbstractTimestamp
+
+        def value_coerced?(value)
+          value.is_a?(Time) && value.usec % 1000 == 0
+        end
+
+        private
+
+        def coerce_time(value)
+          Time.at(value.to_i, value.usec / 1000 * 1000)
```
jturkel: Same question about timezones.

CHANGELOG.md
tjwp: I think this was prior to me merging with master.

lib/avromatic/model/attribute/timestamp_micros.rb
tjwp: The `Time` value created here is local time. The representation as a number of seconds is always Epoch (UTC).

jturkel:
Looks good. :shipit:

PR #39, opened 2016-10-03
Add hooks to preload models and handle Rails reload

tjwp:
Let me know if you want me to cut a pre-release with this. I'll update the README as a follow-up task. 

Prime: @jturkel


tjwp:
Preview of alerts-service change: https://github.com/salsify/alerts-service/compare/feature/avromatic-reload?expand=1

PR #41, opened 2016-10-24
New approach for extended model that must be loaded first

tjwp:
This change replaces the `Avromatic.on_initialize` proc with an `Avromatic.preregister_models` array that can be used to specify an array of models that are registered at the end of `.configure` and on code reloading for Rails applications.

The previous approach relied on the side-effect of a model being registered when it is defined. This was problematic for the first call to the `to_prepare` block for a Rails application because the registry would be cleared, but the constants were still defined for the models, so they would not be re-add simply by referencing. 

I considered disabling the first call to the `to_prepare` block, or disabling the clearing of the registry for the first call to the `to_prepare` block. But being explicit about registering these models seemed cleaner. The configuration could have been left as a generic proc, but that would have a call to register each of the models, so making the structure explicit and putting the registration into avromatic seemed cleaner.

Obviously this change breaks compatibility for any apps using `on_initialize`.

Prime: @jturkel

### DIFF
--- a/lib/avromatic.rb
+++ b/lib/avromatic.rb
@@ -19,7 +19,7 @@ module Avromatic
 
   def self.configure
     yield self
-    preregister_models!
+    eager_load_models!
   end
 
   def self.build_schema_registry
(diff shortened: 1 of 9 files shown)


tjwp:
@jturkel Ready for another pass

lib/avromatic.rb
```diff
@@ -43,8 +44,19 @@ def self.build_messaging!
 
   def self.prepare!
     nested_models.clear
-    on_initialize.call if on_initialize
+    preregister_models!
   end
+
+  def self.preregister_models=(value)
```
jturkel: Should we use a more descriptive variable name than `value`?

spec/avromatic/model_registry_spec.rb
```diff
@@ -19,9 +18,24 @@
...
+      it "registers the model" do
+        instance.register_if_missing(model)
+        expect(instance.registered?('test.nested_record')).to eql(true)
+      end
+    end
+
+    context "for model that has been registered" do
+      before { instance.register(model) }
 
+      it "does not raise an error" do
```
jturkel: Should we assert that the "old" version of the model is still registered (which may require some mocking)?

lib/avromatic.rb
tjwp: Renamed to `models`. Based on discussion yesterday I'll probably also rename the method to `#eager_load_models`.

tjwp:
* Replaced `preregister` with `eager_load`. 
* Replace `register_if_missing` with a different implementation that provides additional checks.
* Add some comments around registration.

@jturkel Mind taking another look?

PR #42, opened 2016-10-26
Fix Rails initialization again

tjwp:
This change is part of how I originally considered fixing the initialization issues with Rails. Since no code reloading happens between initialize and the first call to `to_prepare` do not clear the nested models registry on the first call.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/railtie.rb
+++ b/lib/avromatic/railtie.rb
@@ -5,6 +5,14 @@ module Avromatic
         config.logger = Rails.logger
       end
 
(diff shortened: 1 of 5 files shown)


lib/avromatic/railtie.rb
```diff
@@ -5,8 +5,11 @@ class Railtie < Rails::Railtie
         config.logger = Rails.logger
       end
 
+      first_prepare = true
```
jturkel: This could use some comments

PR #43, opened 2016-11-04
Fix for optional array and map fields

tjwp:
The "class" that is used for Virtus attributes is not always a Class. For arrays and maps it will be an instance of an Array or Hash. This change avoids performing undefined checks on those values.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/type_registry.rb
+++ b/lib/avromatic/model/type_registry.rb
@@ -39,6 +39,9 @@ module Avromatic
 
       attr_reader :custom_types
 
+      # The type that is used to define a Virtus attribute may be a Class or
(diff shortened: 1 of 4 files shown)


lib/avromatic/model/type_registry.rb
```diff
@@ -38,6 +38,10 @@ def fetch(object, field_class = nil)
       private
 
       attr_reader :custom_types
+
+      def union_attribute?(attribute_type)
+        attribute_type.is_a?(Class) && attribute_type < Avromatic::Model::AttributeType::Union
```
jturkel: This could use a comment

PR #46, opened 2017-01-06
Add interfaces to deserialize as a hash of model attributes

tjwp:
This change exposes the hash of attributes prior to building a model during deserialization.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/model/message_decoder.rb
+++ b/lib/avromatic/model/message_decoder.rb
@@ -47,24 +47,42 @@ module Avromatic
           Avromatic.build_schema_registry
       end
 
-      # If two arguments are specified then the first is interpreted as the
-      # message key and the second is the message value. If there is only one
-      # arg then it is used as the message value.
       # @return [Avromatic model]
       def decode(*args)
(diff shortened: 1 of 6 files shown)


lib/avromatic/model/message_decoder.rb
```diff
@@ -47,24 +47,42 @@ def initialize(*models, schema_registry: nil, registry_url: nil)
...
+
+      # If two arguments are specified then the first is interpreted as the
+      # message key and the second is the message value. If there is only one
+      # arg then it is used as the message value.
+      def with_decode_args(*args)
+        message_key, message_value = extract_key_and_value(*args)
+        model_key = model_key_for_message(message_key, message_value)
         raise UnexpectedKeyError.new(model_key) unless model_map.key?(model_key)
-        model_map[model_key].avro_message_decode(message_key, message_value)
+        yield(model_map[model_key], message_key, message_value)
```
jturkel: Any reason to yield these rather than return them?

tjwp: Nope, I can change it.

PR #47, opened 2017-01-19
Optimize union assignment when building models from Avro messages

tjwp:
This change introduces `Avromatic` subclasses for `AvroTurf::Messaging` and `Avro::IO::DatumReader`. `Avromatic::Messaging` uses the new `Avromatic::IO::DatumReader` to read Avro messages.

`Avromatic::IO::DatumReader` changes the behavior of the standard `DatumReader` to include an additional entry in the hash returned for a record giving the index of corresponding union member.

The `Avromatic` code for assigning to union attributes makes use of this additional member index information if present.

In testing with one of our more complex schemas, this makes model building with unions more than 2x faster.

A new configuration option, `use_custom_datum_reader`, is included and defaults to true. This can be set to false to disable the optimization and the additional entry returned by the `DatumReader`.

Also included here are some updates to the versions of gems that we test against using Appraisals.

Prime: @will89

### DIFF
--- a/spec/avromatic/io/datum_reader_spec.rb
+++ b/spec/avromatic/io/datum_reader_spec.rb
@@ -1,0 +1,53 @@
+describe Avromatic::IO::DatumReader do
+  let(:test_class) do
+    Avromatic::Model.model(schema_name: schema_name)
+  end
+  let(:instance) { test_class.new(values) }
+  let(:avro_message_value) { instance.avro_message_value }
+  let(:attributes) { test_class.avro_message_attributes(avro_message_value) }
+
+  context "primitive types" do
+    let(:schema_name) { 'test.primitive_types' }
+    let(:values) do
+      {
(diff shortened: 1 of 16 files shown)


spec/avromatic/io/datum_reader_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    it "includes the member index in the decoded hash" do
+      expect(attributes['message'][described_class::UNION_MEMBER_INDEX]).to eq(1)
+    end
+
+    context "when use_custom_datum_reader is false" do
+      before do
+        allow(Avromatic).to receive(:use_custom_datum_reader).and_return(false)
+      end
+
+      it "includes the member index in the decoded hash" do
```
will89: I think this should say `excludes`, based on the expectation.

will89:
🚢

PR #48, opened 2017-01-19
Add support for looking up schema by fingerprint before registering

tjwp:
This change adds patch for `AvroTurf::SchemaRegistry` and `FakeSchemaRegistryServer` (both from `AvroTurf`) to support the `GET /subjects/:subject/fingerprints/:fingerprint` endpoint that is used to lookup schemas prior to registering.

Patching these classes seemed like the least intrusive way to add this support. By modifying the `SchemaRegistry` client, the `Messaging` API does not need to be touched.

All of our applications that use Avro use this gem, so they will automatically pick up this additional functionality, and tests use the fake server via `avromatic/rspec` which is updated here to pull in the patched version.

Prime: @will89

### DIFF
--- a/lib/avromatic/schema_registry_patch.rb
+++ b/lib/avromatic/schema_registry_patch.rb
@@ -1,0 +1,23 @@
+require 'avro_turf'
+require 'avro_turf/schema_registry'
+
+AvroTurf::SchemaRegistry.class_eval do
+  # Override register to first check if a schema is registered by fingerprint
+  def register(subject, schema)
+    raise 'schema must be an Avro::Schema' unless schema.is_a?(Avro::Schema)
+
+    registered = false
+    data = begin
+             get("/subjects/#{subject}/fingerprints/#{schema.sha256_fingerprint.to_s(16)}")
+           rescue
+             registered = true
+             post("/subjects/#{subject}/versions", body: { schema: schema.to_s }.to_json)
+           end
+
+    id = data.fetch('id')
+
+    @logger.info("#{registered ? 'Registered' : 'Found'} schema for subject `#{subject}`; id = #{id}")
+
+    id
+  end
+end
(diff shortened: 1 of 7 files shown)


lib/avromatic/schema_registry_patch.rb
```diff
@@ -0,0 +1,23 @@
+require 'avro_turf'
+require 'avro_turf/schema_registry'
+
+AvroTurf::SchemaRegistry.class_eval do
+  # Override register to first check if a schema is registered by fingerprint
+  def register(subject, schema)
+    raise 'schema must be an Avro::Schema' unless schema.is_a?(Avro::Schema)
```
will89: Comparing to this implementation, https://github.com/salsify/avro_turf/blob/salsify-master/lib/avro_turf/schema_registry.rb#L19. Why do we decide to raise an error here?

tjwp: I was thinking that an `Avro::Schema` object was required because we need to get a fingerprint from it for the first request, we always pass a schema object using this library. But if passed a string I guess we could attempt to parse it here.

will89:
Looks good 🚢

lib/avromatic/schema_registry_patch.rb
```diff
@@ -0,0 +1,27 @@
...
+  def register(subject, schema)
+    schema_object = if schema.is_a?(String)
+                      Avro::Schema.parse(schema)
+                    else
+                      schema
+                    end
+
+    registered = false
+    data = begin
+             get("/subjects/#{subject}/fingerprints/#{schema_object.sha256_fingerprint.to_s(16)}")
```
jturkel: Should we provide a configuration option to opt out of this for anyone using the Confluent schema registry? Looks like the code would still work but will make an extra network request that will always fail when resolving a schema.

PR #51, opened 2017-02-01
Refactor to expose AvroTurf::SchemaRegistry#lookup_subject_schema

tjwp:
I thought it made sense to support the additional endpoint directly in the `SchemaRegistry` client, so this change moves things around in the patch to expose it.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/schema_registry_patch.rb
+++ b/lib/avromatic/schema_registry_patch.rb
@@ -9,7 +9,7 @@ module Avromatic
 
       begin
         lookup_subject_schema(subject, schema)
-      rescue
+      rescue Excon::Errors::NotFound
         data = post("/subjects/#{subject}/versions", body: { schema: schema.to_s }.to_json)
         id = data.fetch('id')
         @logger.info("Registered schema for subject `#{subject}`; id = #{id}")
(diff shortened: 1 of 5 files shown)


lib/avromatic/schema_registry_patch.rb
```diff
@@ -7,24 +7,26 @@ module CacheableSchemaRegistration
     def register(subject, schema)
       return super unless Avromatic.use_cacheable_schema_registration
 
+      begin
+        lookup_subject_schema(subject, schema)
+      rescue
```
jturkel: Not new but should this only rescue `Excon::Errors::NotFound`?

tjwp: I intentionally left this rescuing all errors so that it would work even against a server that does not support the endpoint.

jturkel: Wouldn't a server that doesn't support this endpoint return a 404? As written it seems like we'll mask bugs in this code.

tjwp: Fair enough, I verified that our server returns a 404 for an unsupported endpoint, and I tested an instance of the Confluent Schema Registry and it also returns a 404 for this request.

PR #52, opened 2017-02-02
Add register_schemas! method to generated models

tjwp:
This method provides a convenient way to register the schemas associated with a model without needing to publish a message.

Eventually, this could be surrounded by calls that alter the compatibility level for a subject in the schema registry, if necessary. Currently the schema registry client doesn't support the Config API so I'm adding this by itself for now.

Prime: @jturkel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # avromatic changelog
 
+## v0.17.0
+- Add `.register_schemas!` method to generated models to register the associated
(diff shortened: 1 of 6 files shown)


jturkel:
It would be awesome to add a similar to method to salsify_kafka that would register schemas for all actor events + `Avromatic.eager_load_models`.

tjwp:
Should be easier to add that `salsify_kafka` method now that there's this to build on!

PR #54, opened 2017-02-17
Namespace Avro errors in Avromatic::IO::DatumReader

tjwp:
Added some test coverage to demonstrate that correct errors are now raised.

Prime: @mattcross


mattcross:
🚢 Thanks!

PR #61, opened 2017-04-06
Remove monkey-patches and use avro_schema_registry-client instead 

tjwp:
Prime: @jturkel


lib/avromatic.rb
```diff
@@ -27,9 +27,15 @@ def self.configure
 
   def self.build_schema_registry
     raise 'Avromatic must be configured with a registry_url' unless registry_url
-    AvroTurf::CachedConfluentSchemaRegistry.new(
-      AvroTurf::ConfluentSchemaRegistry.new(registry_url, logger: logger)
-    )
+    if use_cacheable_schema_registration
```
jturkel: Is there a better name for this setting to disambiguate the two forms of caching (local vs. reverse proxy)? It's a bit confusing to see caching in both branches of the conditional.

tjwp: `use_http_cacheable_schema_registration` or `use_schema_fingerprint_lookup`?

jturkel: I vote for `use_schema_fingerprint_lookup` in case other schema registries introduce other forms of caching.

tjwp:
@jturkel Updated this to use `use_schema_fingerprint_lookup`.

PR #63, opened 2017-04-12
Fix optional union decoding

tjwp:
This fixes the issue that we saw with optional union decoding including the index of the null type.

In the process of adding some test coverage around this I found another problematic case around null as a union member. That one is harder to fix, and not something that I'd expect us to do. For now I've marked the specs as pending for that case, and added an exception when we detect the situation.

Prime: @will89

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # avromatic changelog
 
+## v0.21.1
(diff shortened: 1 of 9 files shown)


will89:
Looks good to me 🚢 🛳 :shipit:

PR #66, opened 2017-06-01
Allow mutable models via config option

Donavan:
This PR adds a new configuration option to control the immutability of generated models.  Models are immutable by default.  Passing immutable: false to the builder will include Virtus.model instead of Virtus.value_object


tjwp:
@Donavan Thanks for the PR. I implemented the new option as `mutable: true` in https://github.com/salsify/avromatic/pull/67 rather than requesting a bunch of changes here. Hopefully this meets your needs. I've just released v0.23.0.

PR #69, opened 2017-06-08
Add Avromatic::IO::DatumWriter to optimize union encoding

tjwp:
This change adds the index of the union member type to records within unions. When this information is present the new `DatumWriter` class can encode the union datum directly without having to check the other member types.

Prime: @jturkel

### DIFF
--- a/lib/avromatic/messaging.rb
+++ b/lib/avromatic/messaging.rb
@@ -39,9 +39,6 @@ module Avromatic
       schema_id = @registry.register(subject || schema.fullname, schema)
 
       stream = StringIO.new
-
-      # The following line differs from the parent class to use a custom DatumWriter
-      writer = Avromatic::IO::DatumWriter.new(schema)
       encoder = Avro::IO::BinaryEncoder.new(stream)
 
       # Always start with the magic byte.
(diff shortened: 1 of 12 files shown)


lib/avromatic/messaging.rb
```diff
@@ -1,8 +1,8 @@
 require 'avro_turf/messaging'
-require 'avromatic/io/datum_reader'
+require 'avromatic/io'
 
 module Avromatic
-  # Subclass AvroTurf::Messaging to use a custom DatumReader for decode.
+  # Subclass AvroTurf::Messaging to use a custom DatumReader and DatumWriter
```
jturkel: We're overriding everything except `initialize` now :)

lib/avromatic/messaging.rb
```diff
@@ -30,5 +30,30 @@ def decode(data, schema_name: nil, namespace: @namespace)
...
+      schema = @schema_store.find(schema_name, namespace)
+
+      # Schemas are registered under the full name of the top level Avro record
+      # type, or `subject` if it's provided.
+      schema_id = @registry.register(subject || schema.fullname, schema)
+
+      stream = StringIO.new
+
+      # The following line differs from the parent class to use a custom DatumWriter
+      writer = Avromatic::IO::DatumWriter.new(schema)
```
jturkel: nit: can we declare the `writer` closer to where we first use it?

lib/avromatic/messaging.rb
tjwp: I know :( At least it shows we're using the same interface!

lib/avromatic/messaging.rb
tjwp: I can. It's in its current location to more closely match the `avro_turf` source, but it doesn't matter much.

PR #72, opened 2017-07-24
Memoize value_attributes_for_avro

will89:
In connection with this change https://github.com/salsify/dandelion/pull/8648, this allows reuse of a product events value_attributes_for_avro. This is useful for when products match multiple subscriptions at once.

Hopefully this is a cacheable value :crossed_fingers: 

Before
screen shot 2017-07-23 at 11 42 43 am

After
screen shot 2017-07-23 at 11 43 10 am

prime: @tjwp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # avromatic changelog
 
 ## v0.26.0
-- Caches result of Avromatic::Model::RawSerialization#value_attributes_for_avro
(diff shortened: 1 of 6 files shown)


tjwp:
By default Avromatic models are Virtus value objects so this memoization is reasonable in that case. But it probably should be disable if a model is defined as mutable. Or something would be needed to clear the memoized value on modification.

PR #73, opened 2017-08-01
Virtus::ValueObject allowed_writer_methods memoization

will89:
This PR allows the Virtus::ValueObject to reuse the the result of `#allowed_writer_methods`. This method gets called a lot when instantiating a new avromatic model.

```
products = Product.order(:external_id).limit(100)
# Work out the db reference resolving
products.each { |p| Core::Events::Product.build(p); nil }; nil

Benchmark.ips { |x| x.report('foo') {  products.each { |p| Core::Events::Product.build(p); nil }; nil; nil } }
master
Warming up --------------------------------------
                 foo     1.000  i/100ms
                 foo      0.396  (± 0.0%) i/s -      2.000  in   5.062918s

virtus patch
Warming up --------------------------------------
                 foo     1.000  i/100ms
                 foo      1.456  (± 0.0%) i/s -      8.000  in   5.523601s
```

Before FlameGraph
screen shot 2017-08-01 at 9 25 55 am

After FlameGraph
screen shot 2017-08-01 at 9 26 44 am

prime: @tjwp


tjwp:
Discussed offline, but it would be better to limit this patch to immutable Avromatic models by including an additional module in the builder.

jturkel:
Out of curiosity is the memoization in Virtus not working properly? https://github.com/solnic/virtus/blob/68343e16f61644196b0a346713b46a8b29e7c2cf/lib/virtus/value_object.rb#L89

will89:
I believe each instance performs its own memoization of those methods instead it being shared across instances.
Given a model class `MyModel`
instance1 = MyModel.new
instance2 = MyModel.new

Without this
instance1.allowed_writer_methods.object_id will not equal instance2.allowed_writer_methods.object_id

jturkel:
Hmm. It looks like those methods are added at the class level: https://github.com/solnic/virtus/blob/68343e16f61644196b0a346713b46a8b29e7c2cf/lib/virtus/value_object.rb#L40.

will89:
Oddly enough, I think `Virtus.model` already memoizes this value.

lib/avromatic/model/allowed_writer_methods_memoization.rb
```diff
@@ -0,0 +1,22 @@
+require 'active_support'
```
tjwp: Is this necessary?

lib/avromatic/model/allowed_writer_methods_memoization.rb
```diff
@@ -0,0 +1,22 @@
...
+  module Model
+    module AllowedWriterMethodsMemoization
+
+      def self.prepended(base)
+        base.prepend(ClassMethods)
+      end
+
+      def self.included(base)
+        base.class_attribute :virtus_object_allowed_writer_methods
+      end
```
tjwp: Can you do the `base.prepend` here?

spec/avromatic/model/allowed_writer_methods_memoization_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        expect(instance1.allowed_writer_methods.object_id).to eq(instance2.allowed_writer_methods.object_id)
+      end
+    end
+
+    context "mutable model" do
+      let(:test_class) { Avromatic::Model.model(value_schema_name: 'test.encode_value', mutable: true) }
+      let(:instance1) { test_class.new(str1: 'a', str2: 'b') }
+      let(:instance2) { test_class.new(str1: 'c', str2: 'd') }
+
+      it "does not share the return value across instances" do
```
tjwp: Update desc.

lib/avromatic/model/allowed_writer_methods_memoization.rb
```diff
@@ -0,0 +1,16 @@
+module Avromatic
+  module Model
+    module AllowedWriterMethodsMemoization
+      def self.included(base)
+        base.class_attribute :virtus_object_allowed_writer_methods
+        base.prepend(ClassMethods)
+      end
+
+      module ClassMethods
```
tjwp: I think we should rename this `InstanceMethods`. The module that is being prepended is overriding instance methods. After that I think this is good to go.

PR #75, opened 2017-08-17
Cache avro encoding for immutable models [IN-48]

jkapell:
This change introduces caching of avro encodings for immutable models with an eye toward speeding up percolate match handling when products match multiple subscriptions.

RawSerialization::recursive_serialize was changed s.t. for immutable models, instead of producing a hash of serialized values, it provides a signal to the DatumWriter to delegate back to the model for the avro encoding. The DatumWriter was correspondingly updated to respond to this hint by calling the model's avro_raw_value which is now memoized.

Before
image

After
image

prime: @tjwp

### DIFF
--- a/lib/avromatic/io/datum_writer.rb
+++ b/lib/avromatic/io/datum_writer.rb
@@ -20,6 +20,14 @@ module Avromatic
         encoder.write_long(index_of_schema)
         write_data(writers_schema.schemas[index_of_schema], datum, encoder)
       end
+
+      def write_record(writers_schema, datum, encoder)
+        if datum.is_a?(Hash) && datum.key?(Avromatic::IO::ENCODING_PROVIDER)
+          encoder.write(datum[Avromatic::IO::ENCODING_PROVIDER].avro_raw_value)
+        else
+          super
+        end
+      end
     end
   end
 end
--- a/lib/avromatic/model/raw_serialization.rb
+++ b/lib/avromatic/model/raw_serialization.rb
@@ -29,6 +29,9 @@ module Avromatic
             if value.is_a?(Avromatic::Model::Attributes)
               hash = value.value_attributes_for_avro
               if Avromatic.use_custom_datum_writer
+                unless value.class.config.mutable
+                  hash = { Avromatic::IO::ENCODING_PROVIDER => value }
+                end
                 member_index = member_types.index(value.class) if member_types.any?
                 hash[Avromatic::IO::UNION_MEMBER_INDEX] = member_index if member_index
               end
@@ -67,7 +70,11 @@ module Avromatic
         end
 
         def avro_raw_value
-          avro_raw_encode(value_attributes_for_avro, :value)
+          if self.class.config.mutable
+            avro_raw_encode(value_attributes_for_avro, :value)
+          else
+            @avro_raw_value ||= avro_raw_encode(value_attributes_for_avro, :value)
+          end
         end
 
         def avro_raw_key
--- a/lib/avromatic/model/raw_serialization.rb
+++ b/lib/avromatic/model/raw_serialization.rb
@@ -30,6 +30,8 @@ module Avromatic
               hash = value.value_attributes_for_avro
               if Avromatic.use_custom_datum_writer
                 unless value.class.config.mutable
+                  # n.b. Ideally we'd just return value here instead of wrapping it in a
+                  # hash but then we'd have no place to stash the union member index...
                   hash = { Avromatic::IO::ENCODING_PROVIDER => value }
                 end
                 member_index = member_types.index(value.class) if member_types.any?
--- a/spec/avromatic/io/datum_writer_spec.rb
+++ b/spec/avromatic/io/datum_writer_spec.rb
@@ -50,4 +50,30 @@ describe Avromatic::IO::DatumWriter do
       end
     end
   end
+
+  describe "#write_record" do
+    let(:message) { datum }
+
+    before do
+      allow(message).to receive(:avro_raw_value)
+      allow(encoder).to receive(:write)
+      datum_writer.write_record(union_schema.schemas[1], datum, encoder)
+    end
+
+    context "when the datum includes an encoding provider" do
+      let(:message) { datum[Avromatic::IO::ENCODING_PROVIDER] }
+
+      it "delegates encoding to the model" do
+        expect(message).to have_received(:avro_raw_value)
+      end
+    end
+
+    context "when the datum doesn't include an encoding provider" do
+      let(:use_custom_datum_writer) { false }
+
+      it "doesn't delegate encoding to the model" do
+        expect(message).not_to have_received(:avro_raw_value)
+      end
+    end
+  end
 end
--- a/spec/avromatic/io/datum_writer_spec.rb
+++ b/spec/avromatic/io/datum_writer_spec.rb
@@ -53,9 +53,10 @@ describe Avromatic::IO::DatumWriter do
 
   describe "#write_record" do
     let(:message) { datum }
+    let(:pre_encoded) { 'foo' }
 
     before do
-      allow(message).to receive(:avro_raw_value)
+      allow(message).to receive(:avro_raw_value).and_return(pre_encoded)
       allow(encoder).to receive(:write)
       datum_writer.write_record(union_schema.schemas[1], datum, encoder)
     end
@@ -64,7 +65,7 @@ describe Avromatic::IO::DatumWriter do
       let(:message) { datum[Avromatic::IO::ENCODING_PROVIDER] }
 
       it "delegates encoding to the model" do
-        expect(message).to have_received(:avro_raw_value)
+        expect(encoder).to have_received(:write).with(pre_encoded)
       end
     end
 
@@ -72,7 +73,9 @@ describe Avromatic::IO::DatumWriter do
       let(:use_custom_datum_writer) { false }
 
       it "doesn't delegate encoding to the model" do
-        expect(message).not_to have_received(:avro_raw_value)
+        union_schema.schemas[1].fields.each do |field|
+          expect(datum_writer).to have_received(:write_data).with(field.type, datum[field.name], encoder)
+        end
       end
     end
   end
--- a/spec/avromatic/model/raw_serialization_spec.rb
+++ b/spec/avromatic/model/raw_serialization_spec.rb
@@ -3,6 +3,7 @@ require 'avro/builder'
 
 describe Avromatic::Model::RawSerialization do
   let(:values) { { id: rand(99) } }
+  let(:test_class) { Avromatic::Model.model(value_schema_name: schema_name) }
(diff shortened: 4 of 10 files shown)


spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+      end
+      let(:sub) { Avromatic.nested_models['test.__nested_record_sub_record'].new(str: 'b', i: 1) }
+      let(:values) { { str: 'a', sub: sub } }
+
+      it "reuses cacheable attributes" do
+        expected = values.deep_stringify_keys
+        expected['sub'] = { encoding_provider => sub }
+        actual = instance.value_attributes_for_avro
+        expect(actual).to eq(expected)
+        expect(actual['sub'][encoding_provider].object_id).to eq(sub.object_id)
```
tjwp: I think another way to write this expectation would be to use `equal`:

```ruby
expect(actual['sub'][encoding_provider]).to equal(sub)
```

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+      end
+    end
+
+    context "with repeated references to a named type" do
+      let!(:test_class) do
```
tjwp: Does this need to be `let!`?

spec/avromatic/io/datum_writer_spec.rb
```diff
@@ -50,4 +50,30 @@
...
+    end
+
+    context "when the datum doesn't include an encoding provider" do
+      let(:use_custom_datum_writer) { false }
+
+      it "doesn't delegate encoding to the model" do
+        expect(message).not_to have_received(:avro_raw_value)
+      end
+    end
+  end
```
tjwp: My inclination for these specs would be to set an expectation with the arg that the encoder will receive with write. I.e. without the encoding provider `expect(encoder).to have_received(:write).with(<whatever the hash would be>)`.

Then for the case where the encoding provider is used, stub a result from `:avro_raw_value`:

```ruby
allow(message).to receive(:avro_raw_value).and_return('pre-encoded value')
```

And then expect that encoder receives `:write` with `'pre-encoded value'`.

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+        Avromatic::Model.model(value_schema_name: 'test.wrapper')
+      end
+      let(:wrapped1) { Avromatic.nested_models['test.wrapped1'].new(i: 42) }
+      let(:wrapped2) { Avromatic.nested_models['test.wrapped2'].new(i: 78) }
+      let(:values) { { sub1: wrapped1, sub2: wrapped1, sub3: wrapped2 } }
+
+      it "reuses cacheable attributes" do
+        expected = values.deep_stringify_keys.map { |k, v| [k, { encoding_provider => v }] }.to_h
```
tjwp: I think you can use `transform_values` here, if its available (subject to what is included from activesupport).

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+        actual = instance.value_attributes_for_avro
+        expect(actual).to eq(expected)
+        expect(actual['sub1'][encoding_provider].object_id).to eq(wrapped1.object_id)
```
tjwp: Use `equal` for these instead of referencing `object_id`?

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+        expect(actual['sub2'][encoding_provider].object_id).to eq(wrapped1.object_id)
+        expect(actual['sub3'][encoding_provider].object_id).to eq(wrapped2.object_id)
+      end
+    end
+
+    context "with reference to a mutable attribute" do
+      let!(:test_class) do
+        Avromatic::Model.model(value_schema_name: 'test.wrapper')
+      end
```
tjwp: `let!` again. Also could dry this up by defining something like the following up above:

```ruby
let(:schema_name) { 'test.wrapper' }
let(:test_class) do
  Avromatic::Model.model(value_schema_name: schema_name)
end
```

And just redefine `schema_name`, if necessary, down below.

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+
+      it "doesn't cache mutable attributes" do
+        expected = values.deep_stringify_keys
+        expected['sub1'] = wrapped1.value_attributes_for_avro
+        expected['sub2'] = wrapped2.value_attributes_for_avro
+        expected['sub3'] = { encoding_provider => wrapped3 }
+        actual = instance.value_attributes_for_avro
+        expect(actual).to eq(expected)
+        expect(actual['sub1'][encoding_provider].object_id).not_to eq(wrapped1.object_id)
```
tjwp: `equal` again?

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
         it "caches a hash of attributes that will be encoded using avro" do
           value_attributes1 = instance.value_attributes_for_avro
           value_attributes2 = instance.value_attributes_for_avro
           expect(value_attributes1.object_id).to eq(value_attributes2.object_id)
         end
+
+        it "caches the avro encoding" do
+          encoded1 = instance.avro_raw_value
+          encoded2 = instance.avro_raw_value
+          expect(encoded1.object_id).to eq(encoded2.object_id)
```
tjwp: `equal`?

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+        it "doesn't cache hash of attributes that will be encoded using avro" do
           value_attributes1 = instance.value_attributes_for_avro
           value_attributes2 = instance.value_attributes_for_avro
           expect(value_attributes1.object_id).not_to eq(value_attributes2.object_id)
         end
+
+        it "doesn't cache the avro encoding" do
+          encoded1 = instance.avro_raw_value
+          encoded2 = instance.avro_raw_value
+          expect(encoded1.object_id).not_to eq(encoded2.object_id)
```
tjwp: `equal`

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
+        end
       end
     end
 
     context "a record with a union" do
-      let(:test_class) do
+      let!(:test_class) do
```
tjwp: I'm surprised this change was necessary.

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
...
       end
 
       it "includes union member index in the hash of attributes" do
         expected = values.deep_stringify_keys
-        expected['message'][member_index] = 1
-        expect(instance.value_attributes_for_avro).to eq(expected)
+        expected['message'] = { encoding_provider => bar_message, member_index => 1 }
+        actual = instance.value_attributes_for_avro
+        expect(actual).to eq(expected)
+        expect(actual['message'][encoding_provider].object_id).to eq(bar_message.object_id)
```
tjwp: `equal`

lib/avromatic/io/datum_writer.rb
```diff
@@ -20,6 +20,14 @@ def write_union(writers_schema, datum, encoder)
...
         write_data(writers_schema.schemas[index_of_schema], datum, encoder)
       end
+
+      def write_record(writers_schema, datum, encoder)
+        if datum.is_a?(Hash) && datum.key?(Avromatic::IO::ENCODING_PROVIDER)
+          encoder.write(datum[Avromatic::IO::ENCODING_PROVIDER].avro_raw_value)
+        else
+          super
+        end
+      end
```
tjwp: I think the other approach here would be to avoid the hash and just check for an Avromatic model:

```ruby
def write_record(writers_schema, datum, encoder)
  if value.is_a?(Avromatic::Model::Attributes)
    encoder.write(datum.avro_raw_value)
  else
    super
  end
end
```

spec/avromatic/model/raw_serialization_spec.rb
```diff
@@ -26,47 +27,125 @@
       expect(instance.value_attributes_for_avro).to eq(expected)
     end
 
+    context "with a nested record" do
+      let!(:test_class) do
```
jkapell: You're right. I didn't realize I could access nested_models directly via a Model.nested_models. Because I was using Avromatic.nested_models, I need let! for force the side effect. I'll fix it here and throughout.

spec/avromatic/model/raw_serialization_spec.rb
jkapell: That would be perfect but unf not avail until 4.2 and we still support 4.1...

spec/avromatic/io/datum_writer_spec.rb
jkapell: I like this idea. It works nicely for the case where an encoding provider is used but not in the other scenario, in which case write_record invokes super. I could, instead, set an expectation on datum_writer to receive write_data. 

`
it "doesn't delegate encoding to the model" do
  union_schema.schemas[1].fields.each do |field|
    expect(datum_writer).to have_received(:write_data).with(field.type, datum[field.name], encoder)
  end
end
`

lib/avromatic/model/raw_serialization.rb
```diff
@@ -29,6 +29,9 @@ def recursive_serialize(value, name: nil, member_types: nil)
             if value.is_a?(Avromatic::Model::Attributes)
               hash = value.value_attributes_for_avro
               if Avromatic.use_custom_datum_writer
+                unless value.class.config.mutable
+                  hash = { Avromatic::IO::ENCODING_PROVIDER => value }
```
jkapell: That's the approach I wanted to take but it was complicated by the way we serialize union members by stashing an index into the hash. In order to avoid wrapping value in a hash we'd need to figure out some other way to plumb the union member index. I think its doable but I'm not convinced it would be any prettier...

tjwp: Makes sense. As discussed offline let's just include a comment on why it makes sense to always return a Hash.

spec/avromatic/model/raw_serialization_spec.rb
tjwp: Make sense. Minor, but this might be slightly cleaner than going from hash to array to hash:

```ruby
values.deep_stringify_keys.each_with_object({}) { |(k, v), hash| hash[k] = { encoding_provider => v } }
```

jkapell:
@tjwp Thanks for the feedback. I've implemented your suggestions. Please take a look. Thanks!

PR #76, opened 2017-08-31
Fix caching of optional fields

jkapell:
Prime: @jturkel

### DIFF
--- a/lib/avromatic.rb
+++ b/lib/avromatic.rb
@@ -11,7 +11,7 @@ module Avromatic
     attr_accessor :schema_registry, :registry_url, :schema_store, :logger,
                   :messaging, :type_registry, :nested_models,
                   :use_custom_datum_reader, :use_custom_datum_writer,
-                  :use_schema_fingerprint_lookup, :use_encoding_providers
+                  :use_schema_fingerprint_lookup
 
     delegate :register_type, to: :type_registry
   end
@@ -22,11 +22,14 @@ module Avromatic
   self.use_custom_datum_reader = true
   self.use_custom_datum_writer = true
   self.use_schema_fingerprint_lookup = true
-  self.use_encoding_providers = !defined?(Avromatic::Patches::SchemaValidatorPatch).nil?
 
   def self.configure
     yield self
     eager_load_models!
+  end
+
+  def self.use_encoding_providers?
+    use_custom_datum_writer && !defined?(Avromatic::Patches::SchemaValidatorPatch).nil?
   end
 
   def self.build_schema_registry
--- a/lib/avromatic.rb
+++ b/lib/avromatic.rb
@@ -4,13 +4,14 @@ require 'avromatic/model'
 require 'avromatic/model_registry'
 require 'avromatic/messaging'
 require 'active_support/core_ext/string/inflections'
+require 'avromatic/patches'
 
 module Avromatic
   class << self
     attr_accessor :schema_registry, :registry_url, :schema_store, :logger,
                   :messaging, :type_registry, :nested_models,
                   :use_custom_datum_reader, :use_custom_datum_writer,
-                  :use_schema_fingerprint_lookup
+                  :use_schema_fingerprint_lookup, :use_encoding_providers
 
     delegate :register_type, to: :type_registry
   end
@@ -21,6 +22,7 @@ module Avromatic
   self.use_custom_datum_reader = true
   self.use_custom_datum_writer = true
   self.use_schema_fingerprint_lookup = true
+  self.use_encoding_providers = !defined?(Avromatic::Patches::SchemaValidatorPatch).nil?
 
   def self.configure
     yield self
--- a/lib/avromatic.rb
+++ b/lib/avromatic.rb
@@ -29,7 +29,7 @@ module Avromatic
   end
 
   def self.use_encoding_providers?
-    use_custom_datum_writer && !defined?(Avromatic::Patches::SchemaValidatorPatch).nil?
+    use_custom_datum_writer && defined?(Avromatic::Patches::SchemaValidatorPatch)
   end
 
   def self.build_schema_registry
--- a/lib/avromatic/model/raw_serialization.rb
+++ b/lib/avromatic/model/raw_serialization.rb
@@ -29,7 +29,7 @@ module Avromatic
             if value.is_a?(Avromatic::Model::Attributes)
               hash = value.value_attributes_for_avro
               if Avromatic.use_custom_datum_writer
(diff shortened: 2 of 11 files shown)


lib/avromatic/model/raw_serialization.rb
```diff
@@ -29,7 +29,7 @@ def recursive_serialize(value, name: nil, member_types: nil)
             if value.is_a?(Avromatic::Model::Attributes)
               hash = value.value_attributes_for_avro
               if Avromatic.use_custom_datum_writer
-                unless value.class.config.mutable
+                if Avromatic.use_encoding_providers && !value.class.config.mutable
```
jturkel: Minor but often times Ruby code has both positive a negative forms of boolean methods e.g. `mutable` and `immutable?` for a slight improvement in readability. Also this lines relates to my earlier question about the relation between `use_custom_datum_writer` and `use_encoding_providers`.

lib/avromatic/patches.rb
```diff
@@ -0,0 +1,4 @@
+if defined?(AvroPatches)
```
jturkel: See earlier statement about gems and load order.

lib/avromatic/patches/schema_validator_patch.rb
```diff
@@ -0,0 +1,16 @@
+module Avromatic
+  module Patches
+    module SchemaValidatorPatch
+      # This method replaces validate_recursive in AvroPatches::LogicalTypes::SchemaValidatorPatch
+      # to enable validating datums that contain an encoding provider.
+      def validate_recursive(expected_schema, logical_datum, path, result, encoded = false)
+        datum = resolve_datum(expected_schema, logical_datum, encoded)
+        if datum.is_a?(Hash) && datum.key?(Avromatic::IO::ENCODING_PROVIDER)
+          return if expected_schema.sha256_resolution_fingerprint ==
```
jturkel: How should we handle the `else` case here? Presumably we don't want to call `super` with a hash containing `Avromatic::IO::ENCODING_PROVIDER`.

spec/avro/dsl/test/optional_record.rb
```diff
@@ -0,0 +1,10 @@
+namespace :test
+
+record :message do
```
jturkel: Are we going to have problems defining `test.message` in both `optional_record.rb` and `optional_array.rb`?

lib/avromatic.rb
```diff
@@ -4,13 +4,14 @@
...
 require 'avromatic/messaging'
 require 'active_support/core_ext/string/inflections'
+require 'avromatic/patches'
 
 module Avromatic
   class << self
     attr_accessor :schema_registry, :registry_url, :schema_store, :logger,
                   :messaging, :type_registry, :nested_models,
                   :use_custom_datum_reader, :use_custom_datum_writer,
-                  :use_schema_fingerprint_lookup
```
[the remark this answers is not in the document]
jkapell: use_encoding_providers actually isn't intended to be configurable or even user visible. Here's whats going on: in order for avromatic to use encoding providers (and thus cache the avro encodings for immutable models) two things need to be true: 
- use_custom_datum_writer must be true because caching is actually implemented, in part, in the custom datum writer.
- the avro-patches gem must be present because, with this PR, the caching implementation depends on the schema validation patches it provides (which we patch further).

Because avromatic doesn't actually require avro-patches and use_custom_datum_writer is user configurable, we need to check both of these conditions during serialization to figure out whether we can include encoding providers in the output.

I added Avromatic.use_encoding_providers just to avoid having to explicitly check for the presence of avro-patches deep in the serialization code. (raw_serialization.rb:32)

lib/avromatic/model/raw_serialization.rb
jkapell: I like this idea in general but in this particular case it strikes me as potentially confusing. My concern is that someone might read value.class.config.immutable? and interpret it as "is the value.class.config object immutable?".

lib/avromatic.rb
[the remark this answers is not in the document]
jturkel: Makes sense. Can we hide this setting somewhere else so it isn't confused with the other settings here that are user configurable?

lib/avromatic/model/raw_serialization.rb
jturkel: Fair point

lib/avromatic/patches/schema_validator_patch.rb
jkapell: The call to super would result in a validation failure anyway which is what we want but I'll raise an error here to fail faster. The error messages aren't meaningful in our case because Avro::Schema.validate traps errors and returns true/false.

spec/avro/dsl/test/optional_record.rb
jkapell: No because we recreate the model registry before each test in spec_helper.rb.

jkapell:
Thanks for the review @jturkel. Just pushed with your feedback.

lib/avromatic.rb
```diff
@@ -27,6 +28,10 @@ def self.configure
     eager_load_models!
   end
 
+  def self.use_encoding_providers?
+    use_custom_datum_writer && !defined?(Avromatic::Patches::SchemaValidatorPatch).nil?
```
jturkel: Can `!defined?(Avromatic::Patches::SchemaValidatorPatch).nil?` be simplified to `defined?(Avromatic::Patches::SchemaValidatorPatch)?`?

jkapell:
Thanks again @jturkel. I bumped the version to 0.28.0 on my first attempt at this. It was pushed to rubygems but then rolled back, so 0.28.0 is still currently unreleased...

tjwp:
Version should be bumped to 0.28.1. I yanked 0.28.0 but it is possible that it was downloaded and cached by somebody before I yanked it.

jturkel:
@tjwp lives!!!

tjwp:
In fact, I just yanked 0.28.0 from rubygems. There is a good chance it is still in Salsify's Artifactory and should also be deleted from there.

lib/avromatic/model/raw_serialization.rb
tjwp: It might make sense to define `mutable?/immutable?` as static methods on the generated classes that delegate to the config. I've come close to doing that a couple of times. And going one step further instance could delegate to a method on the class.

tjwp:
Probably worth cutting a pre-release and doing a full test run with dandelion if that hasn't happened already.

jkapell:
Thanks @tjwp! Dandelion tests passing w/ these changes.

PR #83, opened 2018-05-24
Add support for Rails 5.2

tjwp:
Added Rails 5.2 to the versions that the gem is tested against.

Also bump the patch level to the latest for Ruby 2.4 and 2.5.

At this point support for Rails 4.1 could probably be dropped, but I've left it in for now.

Prime: @jkapell


tjwp:
@jkapell welcome back and thanks for the review!

PR #85, opened 2018-08-02
Fix compatibility with avro-patches v0.4.0

tjwp:
A recent change to avro-patches modified the arguments for a method that this gem patches. The change here adds a patch for that new version, preserving back-compatibility by checking the version of the avro-patches gem.

While making this change, I updated some of the test configuration. Previously I don't think that tests were actually being run in Travis. The `rake spec` command was returning success without actually testing anything.

I've dropped the testing with ruby 2.2 since that version is no longer receiving updates.

Additionally, I've dropped Rails 4.x from the configurations tested.

Rubocop was also updated to recognize ruby 2.5, and cop offenses with the version have been created.

Test coverage dropped because there are now separate patches used depending on the avro-patches version, so I've simplified that part of the test setup.

Prime: @jkapell


tjwp:
@jkapell do you have some time to look at this or would you like to nominate somebody else?

rbrowneupstart:
Any way I can help move this PR/issue along? This is a nice gem that would be great to have up to date with avro-patches.

tjwp:
I've released v0.33.0 with this fix.

PR #88, opened 2018-09-20
Update README.md with missing paren

jturkel:


jturkel:
Indeed. This one required a lot of mental energy.

PR #97, opened 2018-10-09
Better coercion error messages

### LINKED ISSUE #93: [2.0] Provider more context in coercion errors
Coercion errors include details on the value and the target type (e.g. `Avromatic::Model::CoercionError: Could not coerce '1' to a String`) but they don't provide context on where these attributes appear in a model which makes tracking down errors difficult.

jturkel:
This PR improves coercion error messages by giving more context on the location of the coercion error. Here are some examples:

```
# Primitive coercion errors
# Before
Could not coerce '1' to a String
# After
Could not coerce '1' to string when setting Core::Events::DigitalAsset#url

# Custom type coercion errors
# Before
Could not coerce '1' to a SalsifyUuid: Value is not a valid Salsify UUID: '1'
# After
Could not coerce '1' to SalsifyUuid when setting Core::Events::Product#id

# Array coercion errors
# Before
Could not coerce '"foobar"' to a SalsifyUuid: Value is not a valid Salsify UUID: 'foobar'
# After
Could not coerce '["foobar"]' to array[SalsifyUuid] when setting ComputedProperty#formula_reference_property_ids

# Union coercion errors
# Before
Could not coerce '{:id=>1}' to a union of ["Core::Events::Product", "Core::Events::DigitalAsset"]
# After
Could not coerce '{:id=>1}' to union[Core::Events::Product, Core::Events::DigitalAsset] when setting Subscriptions::EntityChangedEvent#entity

# Map coercion errors
# Before
Could not coerce '1' to a String
# After
Could not coerce '{:id=>1}' to map[string => string] when setting Subscriptions::EntityChangedEvent#subscription_correlations
```
Note exceptions will still be properly chained for scenarios where there's an underlying exception (e.g custom type coercion).

Fixes #93.

@tjwp - you're prime
/cc @jkapell

### DIFF
--- a/lib/avromatic/model/types/abstract_timestamp_type.rb
+++ b/lib/avromatic/model/types/abstract_timestamp_type.rb
@@ -14,7 +14,7 @@ module Avromatic
           elsif input.is_a?(::Time) || input.is_a?(::DateTime)
             coerce_time(input)
           else
-            raise Avromatic::Model::CoercionError.new("Could not coerce '#{input.inspect}' to #{self.class.name.demodulize}")
+            raise ArgumentError.new("Could not coerce '#{input.inspect}' to #{name}")
           end
         end
 
--- a/lib/avromatic/model/types/boolean_type.rb
+++ b/lib/avromatic/model/types/boolean_type.rb
@@ -9,7 +9,7 @@ module Avromatic
         end
 
         def name
-          'boolean'
+          'boolean'.freeze
         end
 
         def coerce(input)
--- a/lib/avromatic/model/types/boolean_type.rb
+++ b/lib/avromatic/model/types/boolean_type.rb
@@ -8,11 +8,15 @@ module Avromatic
           VALUE_CLASSES
         end
 
+        def name
+          'boolean'
+        end
+
         def coerce(input)
           if coercible?(input)
             input
           else
-            raise Avromatic::Model::CoercionError.new("Could not coerce '#{input.inspect}' to a Boolean")
+            raise ArgumentError.new("Could not coerce '#{input.inspect}' to #{name}")
           end
         end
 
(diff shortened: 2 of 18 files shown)


lib/avromatic/model/types/abstract_timestamp_type.rb
```diff
@@ -14,7 +14,7 @@ def coerce(input)
           elsif input.is_a?(::Time) || input.is_a?(::DateTime)
             coerce_time(input)
           else
-            raise Avromatic::Model::CoercionError.new("Could not coerce '#{input.inspect}' to #{self.class.name.demodulize}")
+            raise ArgumentError.new("Could not coerce '#{input.inspect}' to #{name}")
```
tjwp: Is it necessary to interpolate values and specify an error message here, since the error will always be rescued and replaced by a `CoercionError` that does not use the original error message?

lib/avromatic/model/types/union_type.rb
```diff
@@ -12,6 +12,10 @@ def initialize(member_types:)
           @value_classes = member_types.flat_map(&:value_classes)
         end
 
+        def name
+          "union[#{member_types.map(&:name).join(', ')}]"
```
tjwp: Is it worth memoizing the non-constant name methods?

lib/avromatic/model/types/boolean_type.rb
```diff
@@ -8,11 +8,15 @@ def value_classes
           VALUE_CLASSES
         end
 
+        def name
+          'boolean'
```
tjwp: Should these constant name values be frozen?

lib/avromatic/model/types/union_type.rb
jturkel: I didn't bother since this method isn't called for a hot codepath. We can always add it later if it becomes a bottleneck.

lib/avromatic/model/types/boolean_type.rb
jturkel: I'll freeze them but I'll also put up a separate PR to add `# frozen_string_literal: true` pragmas to this project too.

PR #98, opened 2018-10-10
Validate required fields during Avro serialization

jturkel:
This PR adds automatic validation of required fields when serializing a model to Avro resulting in much better error messages:

```
# Old exception thrown during serialization
Avro::IO::AvroTypeError:
  The datum nil is not an example of schema {"type":"map","values":"string"}

# Old error messages generated when clients explicitly called valid? or invalid?
Subscription can't be blank
Entity can't be blank
Subscription correlations can't be nil
Entity .property_value_collections [0].property can't be blank

# New exception with improved error message
Avromatic::Model::ValidationError:
  Subscriptions::EntityChangedEvent({:system_message_timestamp=>2018-10-10 16:20:47 UTC, :subscription_id=>nil, :entity_id=>nil, :organization_id=>SalsifyUuid(s-00000001-0000-0000-0000-000000000000), :system_message_id=>SalsifyUuid(s-bcca3502-ae80-4aea-a2b9-104fa9035b18), :subscription_correlations=>nil, :entity=>#<Core::Events::Product id: SalsifyUuid(s-00000002-0000-0000-0000-000000000000), type: "products", external_id: "product_1", property_value_collections: [#<Core::Events::StringEncodedPropertyValueCollection property: nil, locale_id: nil, value_data_type: "string", values: ["foo"]>], created_at: 2018-10-10 12:20:47 -0400, updated_at: 2018-10-10 12:20:47 -0400, destroyed_at: nil, name: "Product 1", thumbnail_url: nil, lock_version: 1, profile_asset: nil, parent: nil>}) cannot be serialized because the following attributes are nil: subscription_id, entity_id, subscription_correlations, entity.property_value_collections[0].property
```

The implementation does not rely on ActiveModel::Validations because:

1) It added performance overhead. The version with "native" validations has 1.5x more throughput for a benchmark creating model instances and serializing them to Avro
2) The error messages are better because ActiveModel's humanization of error strings resulted in a confusing translation back to the underlying schema

Models do still include the `ActiveModel::Validations` module for clients that want to explicitly call `valid?` or `invalid?`.

/cc @jkapell


jturkel:
@tjwp - There might be something going on with Github but I don't see your comment.

lib/avromatic/model/validation.rb
```diff
@@ -2,52 +2,71 @@ module Avromatic
...
+          end
+          results
+        elsif value.is_a?(Hash)
+          results = []
+          value.each do |key, element|
+            nested_results = missing_nested_attributes("#{attribute}['#{key}']", element)
+            results.concat(nested_results)
+          end
+          results
+        elsif value.is_a?(Avromatic::Model::Validation)
```
tjwp: I think the comment I made was that this condition was surprising/potentially confusing. I assume that this is checking for an Avromatic model. Is this a condition that is used elsewhere in the code? Or is there something that would be more intention revealing?

jturkel: This condition doesn't appear anywhere else. Do you think it would be cleaner to check `respond_to?(:missing_avro_attributes)`?

tjwp: I think the condition that I've used for similar purposes elsewhere is `value.is_a?(Avromatic::Model::Attributes)`. I don't know if that is any better

Another option may be to create a helper method here, `model_instance?(value)`, and then define `#model_instance?` using the existing expression or one of the other suggestions here.

jturkel: I agree that `value.is_a?(Avromatic::Model::Attributes)` isn't much clearer :) I think I'll just use `value.respond_to?(:missing_avro_attributes)` since that accurately captures the contract that this branch of the conditional actually cares about.

PR #100, opened 2018-10-11
Additional test of union coercion algorithm

### LINKED ISSUE #96: [2.0] Union coercion algorithm should consider all attributes of input for more accurate matching
This will likely fall out naturally from #92 but we should add tests to confirm

jturkel:
This PR adds a test to verify that the union coercion algorithm selects the first union member type with all of the specified attribute values with the correct types.

Fixes #96

### DIFF
--- a/spec/avromatic/model/builder_spec.rb
+++ b/spec/avromatic/model/builder_spec.rb
@@ -794,6 +794,43 @@ describe Avromatic::Model::Builder do
       end
     end
 
+    context "union of records with overlapping fields" do
+      let(:schema) do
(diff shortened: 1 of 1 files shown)


jturkel:
You can assign to `sub4` if you specify a value (including `nil`) for `c`.

tjwp:
Ah, because the additional `c` attribute will cause it to be rejected for `sub3` by default?

jturkel:
Exactly (which is also why `sub1` doesn't match when specifying `a` and `b`)

PR #104, opened 2018-10-18
Define <attribute>? readers for booleans

jturkel:
This defines readers with the `?` suffix for boolean fields.


tjwp:
Worth mentioning these readers in the README and highlighting that they return false for null?

jturkel:
Yeah. It was one of those undocumented things that we got for free with Virtus. I'll beef up an example with some usage + comments.

PR #105, opened 2018-10-18
Improve type coercion errors

jturkel:
This PR improves type coercion errors by providing more details on the cause of the error. The test changes are the easiest way to see the updated error messages. It also fixes a bug that caused the unknown attribute detection code to generate an incorrect message for string attribute keys.

@skarger - you're prime

### DIFF
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -53,6 +53,9 @@ module Avromatic
             raise Avromatic::Model::CoercionError.new("Value for #{owner.name}##{name} could not be coerced to a #{type.name} " \
               "because a #{input.class.name} was provided but expected a #{type.input_classes.map(&:name).to_sentence(two_words_connector: ' or ', last_word_connector: ', or ')}. " \
               "Provided argument: #{input.inspect}")
+          elsif input.is_a?(Hash) && type.is_a?(Avromatic::Model::Types::UnionType)
+            raise Avromatic::Model::CoercionError.new("Value for #{owner.name}##{name} could not be coerced to a #{type.name} " \
+              "because no union member type has all of the provided attributes: #{input.inspect}")
           else
             raise Avromatic::Model::CoercionError.new("Value for #{owner.name}##{name} could not be coerced to a #{type.name}. " \
               "Provided argument: #{input.inspect}")
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -46,7 +46,8 @@ module Avromatic
           type.coerce(input)
         rescue Avromatic::Model::UnknownAttributeError => e
           raise Avromatic::Model::CoercionError.new("Value for #{owner.name}##{name} could not be coerced to a #{type.name} " \
-            "because the following unexpected attributes were provided: #{e.attributes.join(', ')}. " \
+            "because the following unexpected attributes were provided: #{e.unknown_attributes.join(', ')}. " \
+            "Only the following attributes are allowed: #{e.allowed_attributes.join(', ')}. " \
             "Provided argument: #{input.inspect}")
         rescue StandardError
           if type.input_classes && type.input_classes.none? { |input_class| input.is_a?(input_class) }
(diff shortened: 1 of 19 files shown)


lib/avromatic/model/attributes.rb
```diff
@@ -73,7 +87,10 @@ def initialize(data = {})
         end
 
         unless Avromatic.allow_unknown_attributes || valid_keys.size == data.size
-          raise ArgumentError.new("Unexpected attributes for #{self.class.name}: #{(data.keys - valid_keys).map(&:to_s).join(', ')}. Complete arguments: #{data}")
+          unknown_attributes = data.keys.map(&:to_s) - valid_keys.map(&:to_s)
+          message = "Unexpected arguments for #{self.class.name}#initialize: #{unknown_attributes.join(', ')}. " \
+            "Provided arguments: #{data.inspect}"
+          raise Avromatic::Model::UnknownAttributeError.new(message, unknown_attributes)
```
skarger: Can you also include the expected attributes in the `message` and pass them through to `UnknownAttributeError`?

lib/avromatic/model/attributes.rb
```diff
@@ -44,8 +44,22 @@ def required?
 
         def coerce(input)
           type.coerce(input)
+        rescue Avromatic::Model::UnknownAttributeError => e
+          raise Avromatic::Model::CoercionError.new("Value for #{owner.name}##{name} could not be coerced to a #{type.name} " \
+            "because the following unexpected attributes were provided: #{e.attributes.join(', ')}. " \
+            "Provided argument: #{input.inspect}")
```
skarger: See comment below - it would be helpful to print the expected attributes after the unexpected ones, before the provided argument.

lib/avromatic/model/attributes.rb
```diff
@@ -44,8 +44,22 @@ def required?
...
         rescue StandardError
-          raise Avromatic::Model::CoercionError.new("Could not coerce '#{input.inspect}' to #{type.name} when setting #{owner.name}##{name}")
+          if type.input_classes && type.input_classes.none? { |input_class| input.is_a?(input_class) }
+            raise Avromatic::Model::CoercionError.new("Value for #{owner.name}##{name} could not be coerced to a #{type.name} " \
+              "because a #{input.class.name} was provided but expected a #{type.input_classes.map(&:name).to_sentence(two_words_connector: ' or ', last_word_connector: ', or ')}. " \
+              "Provided argument: #{input.inspect}")
+          elsif input.is_a?(Hash) && type.is_a?(Avromatic::Model::Types::UnionType)
+            raise Avromatic::Model::CoercionError.new("Value for #{owner.name}##{name} could not be coerced to a #{type.name} " \
+              "because no union member type has all of the provided attributes: #{input.inspect}")
```
skarger: If it's straightforward to print the union member types' attributes, that would be helpful for comparison with the provided attributes.

lib/avromatic/model/types/abstract_timestamp_type.rb
```diff
@@ -1,15 +1,22 @@
...
         VALUE_CLASSES = [::Time].freeze
+        INPUT_CLASSES = [::Time, ::DateTime, ::ActiveSupport::TimeWithZone].freeze
 
         def value_classes
           VALUE_CLASSES
         end
 
+        def input_classes
+          INPUT_CLASSES
+        end
```
skarger: Is there a pattern to when you define this as a method and when you set it as an instance variable in `initialize`? I'm wondering if you can standardize the approach (and if that's helpful).

jturkel: I set it in a variable in `initialize` if it's based on other attributes of the instance. I use a constant + reader method If it's static data. We could definitely standardize on the former but I don't see much benefit in that as long as the type classes abide by the same API contract. I'm happy to change it if you feel strongly about it though.

skarger: I don't feel that strongly about it.

lib/avromatic/model/attributes.rb
jturkel: We talked about this offline but I'm going to punt on doing this as part of this PR.

jturkel:
@skarger - ready for another review

PR #108, opened 2018-10-23
Allow generated model attribute accessors to be overridden

jturkel:
This PR puts all generated model methods into a module so they can be overridden by subclasses (which Avromatic 1.0 supported).

### DIFF
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -111,7 +111,7 @@ module Avromatic
       end
 
       module ClassMethods
-        def add_avro_fields
+        def add_avro_fields(generated_methods_module)
           # models are registered in Avromatic.nested_models at this point to
           # ensure that they are available as fields for recursive models.
           register!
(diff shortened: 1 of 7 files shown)


lib/avromatic/model/attributes.rb
```diff
@@ -163,33 +163,25 @@ def define_avro_attributes(schema, allow_optional: true)
...
-              define_method(:clone) { self }
-              define_method(:dup) { self }
+              generated_methods_module.send(:private, "#{field.name}=")
+              generated_methods_module.send(:define_method, :clone) { self }
+              generated_methods_module.send(:define_method, :dup) { self }
             end
           end
         end
-
-        def boolean?(field)
```
jturkel: I moved this to `FieldHelper` and inlined `create_type` to avoid polluting the model class' method namespace.

jturkel:
@kphelps - Pushed a change to invoke setters from model constructors. Only interesting thing was I had to cache the symbolized setter name to avoid regressing the performance of model instantiation. Can you take a look?

PR #109, opened 2018-10-24
Fix timestamp-millis coercion

jturkel:
This fixes a bug where timestamp-millis where not coerced when the number of microseconds was divisible by 1,000 but the number of nanoseconds was not divisible by 1,000,000.

@skarger - you're prime

### DIFF
--- a/spec/avromatic/model/builder_spec.rb
+++ b/spec/avromatic/model/builder_spec.rb
@@ -1044,13 +1044,10 @@ describe Avromatic::Model::Builder do
           end
         end
 
-        it "coerces a time to a union member", check_me_out: true do
-          10_000_000.times do |i|
-            puts i + 1 if (i + 1) % 1000 == 0
-            now = Time.now
-            instance = test_class.new(u: now)
-            expect(instance.u).to eq(Time.at(now.to_i, now.usec))
-          end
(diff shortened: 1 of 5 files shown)


spec/avromatic/model/builder_spec.rb
```diff
@@ -1061,9 +1061,9 @@ def initialize
         end
 
         it "coerces a time to a union member" do
-          now = Time.now
-          instance = test_class.new(u: now)
-          expect(instance.u).to eq(Time.at(now.to_i, now.usec / 1000 * 1000))
+          time = Time.at(1540332724, 123456.789)
+          instance = test_class.new(u: time)
+          expect(instance.u).to eq(Time.at(1540332724, 123_000))
         end
```
skarger: I think to test the bug fix you should use a value like `time = Time.at(1540332724, 123000.789)`. Then you have `time.usec % 1_000 == 0` but `time.nsec % 1_000_000 == 789`. That should fail to coerce with the old code but pass with your change.

With `time = Time.at(1540332724, 123456.789)`,  `time.usec % 1_000 == 456` so the old code would still coerce it.

jturkel:
@skarger - Ready for another review

PR #111, opened 2019-02-11
Reduce string allocations during model initialization

kphelps:
Instead of allocating two strings per attribute, we can cache this on the attribute definition.

Before:
```
                avro   288.272k memsize (     0.000  retained)
                         2.392k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
Warming up --------------------------------------
                avro    40.000  i/100ms
                avro    415.694  (± 5.5%) i/s -      2.080k in   5.019321s
```
After:
```
                avro   259.232k memsize (     0.000  retained)
                         1.666k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
Warming up --------------------------------------
                avro    43.000  i/100ms
                avro    446.097  (± 4.5%) i/s -      2.236k in   5.023264s
```

prime @will89 
cc @jturkel @tjwp

### DIFF
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -29,7 +29,7 @@ module Avromatic
           @field = field
           @type = type
           @name = field.name.to_sym
-          @name_string = field.name.to_s.freeze
+          @name_string = field.name.to_s.dup.freeze
           @setter_name = "#{field.name}=".to_sym
           @default = if field.default == :no_default
                        nil
--- a/lib/avromatic/model/attributes.rb
+++ b/lib/avromatic/model/attributes.rb
@@ -21,7 +21,7 @@ module Avromatic
       end
 
       class AttributeDefinition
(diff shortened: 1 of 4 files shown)


lib/avromatic/model/attributes.rb
```diff
@@ -21,14 +21,15 @@ def initialize(field)
...
-        attr_reader :name, :setter_name, :type, :field, :default, :owner
+        attr_reader :name, :name_string, :setter_name, :type, :field, :default, :owner
         delegate :serialize, to: :type
 
         def initialize(owner:, field:, type:)
           @owner = owner
           @field = field
           @type = type
           @name = field.name.to_sym
+          @name_string = field.name.to_s
```
will89: Is `field.name` guaranteed to be a frozen string? If not should we dup and then freeze it? The real memory allocation savings come from the `initialize` method below right?

PR #118, opened 2019-12-01
Allow for string data time to be coerced 

danhawkins:
In some cases I am passing datetime as a string to the Model, this allows for coercing the string value if possible

### DIFF
--- a/spec/avromatic/model/builder_spec.rb
+++ b/spec/avromatic/model/builder_spec.rb
@@ -336,7 +336,7 @@ describe Avromatic::Model::Builder do
         end
 
         it "coerces an DateTime string to a time with millisecond precision" do
-          time = "2018-10-24 02:12:04.123456 +0400"
+          time = '2018-10-24 02:12:04.123456 +0400'
           instance = test_class.new(ts_msec: time)
           expect(instance.ts_msec).to eq(::Time.at(1540332724, 123_000))
         end
@@ -366,13 +366,13 @@ describe Avromatic::Model::Builder do
         end
 
         it "coerces an DateTime string to a time with microsecond precision" do
-          time = "2018-10-24 02:12:04.123456 +0400"
+          time = '2018-10-24 02:12:04.123456 +0400'
           instance = test_class.new(ts_msec: time)
           expect(instance.ts_msec).to eq(::Time.at(1540332724, 123000))
         end
 
         it "raises an error with bad time string" do
-          time = "bad time string"
+          time = 'bad time string'
           expect { test_class.new(ts_msec: time) }.to raise_error(
             Avromatic::Model::CoercionError,
             'Value for LogicalType#ts_msec could not be coerced to a timestamp-millis. Provided argument: "bad time string"'
--- a/spec/avromatic/model/builder_spec.rb
+++ b/spec/avromatic/model/builder_spec.rb
@@ -341,6 +341,14 @@ describe Avromatic::Model::Builder do
           expect(instance.ts_msec).to eq(::Time.at(1540332724, 123_000))
         end
 
+        it "raises an error with bad time string" do
+          time = 'bad time string'
+          expect { test_class.new(ts_msec: time) }.to raise_error(
+            Avromatic::Model::CoercionError,
+            'Value for LogicalType#ts_msec could not be coerced to a timestamp-millis. Provided argument: "bad time string"'
+          )
+        end
+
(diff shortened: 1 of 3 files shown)


jturkel:
Thanks for opening this PR. I'm hesitant to add support for coercing strings to date times since Avro already supports logical types for date times which provides a type safe way of modeling date time attributes. Would one of these alternatives work for your use case?

1. Override the attributes setter method to perform the necessary coercion
1. Use a [custom type](https://github.com/salsify/avromatic#custom-type-configuration) to override the serialization/deserialization hooks

spec/avromatic/model/builder_spec.rb
```diff
@@ -359,6 +365,20 @@
           expect(instance.ts_usec).to eq(::Time.at(1540332724, 123456))
         end
 
+        it "coerces an DateTime string to a time with microsecond precision" do
+          time = '2018-10-24 02:12:04.123456 +0400'
+          instance = test_class.new(ts_msec: time)
+          expect(instance.ts_msec).to eq(::Time.at(1540332724, 123000))
```
tjwp: Should these lines be:

```
instance = test_class.new(ts_usec: time)
expect(instance.ts_usec).to eq(::Time.at(1540332724, 123456))
```

danhawkins: Ah, yes indeed, that does need a fix

danhawkins:
> Thanks for opening this PR. I'm hesitant to add support for coercing strings to date times since Avro already supports logical types for date times which provides a type safe way of modeling date time attributes. Would one of these alternatives work for your use case?
> 
> 1. Override the attributes setter method to perform the necessary coercion
> 2. Use a [custom type](https://github.com/salsify/avromatic#custom-type-configuration) to override the serialization/deserialization hooks

The reason I am suggesting this is in some cases I need to init the model with json data, in which case you cannot get a datetime in there without messing around.

Given that the initial check for a Time will always prevail over the string, then it should not change the performance or behaviour of the original implementation at all, only when that condition fails will it check the string.

jturkel:
My biggest concern with the proposed change is it doesn't provide enough configurability in terms of how dates are parsed e.g. some clients might want to use iso8601, some may want rfc3339, some may want something else entirely. We've toyed with a more general framework to specify per attribute coercion rules but we haven't seen sufficient need since the typing information typically exists in the schema. I'm definitely open to something like this but I'm not sure exactly what form factor it would have.

In the meantime setters can be overridden to perform the coercion with something like this:

```ruby
class MyModel
  include Avromatic::Model.build(schema_name :my_model)

  def created_at=(value)
    value.is_a?(String) ? super(Time.parse(input)) : super
  end
end
```
If this is a common pattern in your application, you can use some meta-programming to reduce the boilerplate:

```ruby
class BaseModel
  def self.date_fields(*fields)
    fields.each do |field|
      define_method("#{field}=") do |value|
        value.is_a?(String) ? super(Time.parse(input)) : super(value)
      end
    end
  end
end

class MyModel < BaseModel
  include Avromatic::Model.build(schema_name :my_model)

  date_fields :created_at, :updated_at
end
```

danhawkins:
> My biggest concern with the proposed change is it doesn't provide enough configurability in terms of how dates are parsed e.g. some clients might want to use iso8601, some may want rfc3339, some may want something else entirely. We've toyed with a more general framework to specify per attribute coercion rules but we haven't seen sufficient need since the typing information typically exists in the schema. I'm definitely open to something like this but I'm not sure exactly what form factor it would have.
> 
> In the meantime setters can be overridden to perform the coercion with something like this:
> 
> ```ruby
> class MyModel
>   include Avromatic::Model.build(schema_name :my_model)
> 
>   def created_at=(value)
>     value.is_a?(String) ? super(Time.parse(input)) : super
>   end
> end
> ```
> 
> If this is a common pattern in your application, you can use some meta-programming to reduce the boilerplate:
> 
> ```ruby
> class BaseModel
>   def self.date_fields(*fields)
>     fields.each do |field|
>       define_method("#{field}=") do |value|
>         value.is_a?(String) ? super(Time.parse(input)) : super(value)
>       end
>     end
>   end
> end
> 
> class MyModel < BaseModel
>   include Avromatic::Model.build(schema_name :my_model)
> 
>   date_fields :created_at, :updated_at
> end
> ```

OK, makes sense, thanks for the meta approach, I will use that, very clean!

PR #119, opened 2019-12-18
Load the nested models of eager loaded models

jcmfernandes:
Generated nested models register themselves upon being created by
`Avromatic::Model::Builder`. If Rails's autoloader didn't recreate a model that
should be eager loaded, then its nested models won't be registered as we hook on
Railtie `#to_prepare` that itself triggers a clean-up of the nested models
registry.

Therefore, we have to go through the attribute definitions of each eager loaded
model and try to register its nested models.

prime @jturkel

### DIFF
--- a/lib/avromatic.rb
+++ b/lib/avromatic.rb
@@ -86,7 +86,7 @@ module Avromatic
   end
 
   def self.eager_load_models!
-    @eager_load_model_names&.map(&:constantize)&.each(&:register!)
+    @eager_load_model_names&.each { |model_name| model_name.constantize.register! }
   end
   private_class_method :eager_load_models!
 end
--- a/lib/avromatic.rb
+++ b/lib/avromatic.rb
@@ -77,35 +77,46 @@ module Avromatic
         end
       end
     end
-    eager_load_models!(skip_nested_models: skip_clear)
+    eager_load_models!
   end
 
   def self.eager_load_models=(models)
     @eager_load_model_names = Array(models).map { |model| model.is_a?(Class) ? model.name : model }
   end
 
-  def self.eager_load_models!(skip_nested_models: true)
+  def self.eager_load_models!
     (@eager_load_model_names || []).each do |model_name|
       model = model_name.constantize
-      next if !nested_models.ensure_registered_model(model) || skip_nested_models
+      next if nested_models.registered?(model)
 
-      model.attribute_definitions.each do |_attribute, definition|
-        eager_load_nested_models!(definition.type)
+      nested_models.register(model)
+
+      roots = model.attribute_definitions.values.map(&:type)
+      loop do
+        model = roots.shift
+        eager_load_nested_models!(model, roots)
+
+        break if roots.empty?
       end
+
     end
   end
   private_class_method :eager_load_models!
 
-  def self.eager_load_nested_models!(type)
+  def self.eager_load_nested_models!(type, roots)
     case type
     when Avromatic::Model::Types::ArrayType, Avromatic::Model::Types::MapType
       eager_load_nested_models!(type.value_type)
     when Avromatic::Model::Types::CustomType
-      eager_load_nested_models!(type.default_type) # correct?
+      eager_load_nested_models!(type.default_type)
     when Avromatic::Model::Types::UnionType
       type.member_types.each { |member_type| eager_load_nested_models!(member_type) }
     when Avromatic::Model::Types::RecordType
-      nested_models.ensure_registered_model(type.record_class)
+      model = type.record_class
+      unless nested_models.registered?(model)
+        nested_models.register(model)
+        roots.concat(model.attribute_definitions.values.map(&:type))
+      end
     end
   end
   private_class_method :eager_load_nested_models!
--- a/lib/avromatic.rb
+++ b/lib/avromatic.rb
@@ -77,19 +77,38 @@ module Avromatic
         end
       end
     end
-    eager_load_models!
+    eager_load_models!(skip_nested_models: skip_clear)
   end
 
   def self.eager_load_models=(models)
     @eager_load_model_names = Array(models).map { |model| model.is_a?(Class) ? model.name : model }
   end
 
-  def self.eager_load_models!
+  def self.eager_load_models!(skip_nested_models: true)
     (@eager_load_model_names || []).each do |model_name|
-      nested_models.ensure_registered_model(model_name.constantize)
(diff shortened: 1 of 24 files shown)


lib/avromatic.rb
```diff
@@ -77,19 +77,38 @@ def self.prepare!(skip_clear: false)
...
     end
   end
   private_class_method :eager_load_models!
+
+  def self.eager_load_nested_models!(type)
+    case type
+    when Avromatic::Model::Types::ArrayType, Avromatic::Model::Types::MapType
+      eager_load_nested_models!(type.value_type)
+    when Avromatic::Model::Types::CustomType
+      eager_load_nested_models!(type.default_type) # correct?
```
jcmfernandes: @jturkel I assume this makes sense but let me know.

lib/avromatic.rb
```diff
@@ -77,19 +77,38 @@ def self.prepare!(skip_clear: false)
...
+    when Avromatic::Model::Types::UnionType
+      type.member_types.each { |member_type| eager_load_nested_models!(member_type) }
+    when Avromatic::Model::Types::RecordType
+      nested_models.ensure_registered_model(type.record_class)
```
jcmfernandes: Begs the question if we should restart the recursion with the `#attribute_definitions` of `type.record_class`.

lib/avromatic.rb
```diff
@@ -77,19 +77,38 @@ def self.prepare!(skip_clear: false)
...
   def self.eager_load_models=(models)
     @eager_load_model_names = Array(models).map { |model| model.is_a?(Class) ? model.name : model }
   end
 
-  def self.eager_load_models!
+  def self.eager_load_models!(skip_nested_models: true)
     (@eager_load_model_names || []).each do |model_name|
-      nested_models.ensure_registered_model(model_name.constantize)
+      model = model_name.constantize
+      next if !nested_models.ensure_registered_model(model) || skip_nested_models
```
jturkel: Relying on `ensure_registered_model` to return a truthy value is a bit dangerous given the current implementation (i.e. it works today but likely was not intentional). Adding a `next if nested_models.registered?(model)` guard is safer.

lib/avromatic.rb
jturkel: Looks correct to me!

lib/avromatic.rb
```diff
@@ -77,19 +77,38 @@ def self.prepare!(skip_clear: false)
...
+
+      model.attribute_definitions.each do |_attribute, definition|
+        eager_load_nested_models!(definition.type)
+      end
     end
   end
   private_class_method :eager_load_models!
+
+  def self.eager_load_nested_models!(type)
+    case type
```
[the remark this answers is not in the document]
jcmfernandes: Addressed.

lib/avromatic.rb
```diff
@@ -77,19 +77,38 @@ def self.prepare!(skip_clear: false)
         end
       end
     end
-    eager_load_models!
+    eager_load_models!(skip_nested_models: skip_clear)
```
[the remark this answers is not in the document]
jcmfernandes: Addressed,

jcmfernandes:
Hey @jturkel, I did some changes based on your feedback. Let me know what you think. Thanks!

lib/avromatic.rb
```diff
@@ -81,13 +82,11 @@ def self.prepare!(skip_clear: false)
   end
 
   def self.eager_load_models=(models)
-    @eager_load_model_names = Array(models).map { |model| model.is_a?(Class) ? model.name : model }
+    @eager_load_model_names = Array.wrap(models).map { |model| model.is_a?(Class) ? model.name : model }
```
jturkel: Was this change necessary? `Array()` and `Array.wrap()` should behave the same here.

lib/avromatic/model/types/abstract_type.rb
```diff
@@ -31,6 +31,10 @@ def coerced?(_value)
         def serialize(_value, **)
           raise "#{__method__} must be overridden by #{self.class.name}"
         end
+
+        def referenced_models
```
jturkel: Should we call this `referenced_model_classes` to make it more obvious that we're not returning instantiated models?

lib/avromatic/model/types/map_type.rb
```diff
@@ -68,6 +68,10 @@ def serialize(value, strict:)
             end
           end
         end
+
+        def referenced_models
+          [key_type, value_type].tap(&:uniq!).flat_map(&:referenced_models).tap(&:uniq!)
```
jturkel: nit: Doesn't seem like the extra `tap(&:uniq!)` buys us much
```suggestion
          [key_type, value_type].flat_map(&:referenced_models).tap(&:uniq!)
```

lib/avromatic/model/types/boolean_type.rb
```diff
@@ -33,6 +33,10 @@ def coercible?(input)
         def serialize(value, **)
           value
         end
+
+        def referenced_models
+          []
```
jturkel: It might be worth declaring an `EMPTY_ARRAY` constant in `AbstractType` that subclasses can use. I don't expect a noticeable perf gain though since this method won't be called frequently. Up to you.

lib/avromatic/model/types/map_type.rb
jcmfernandes: It's my lazy way of performing the search twice for the same type. I'll leave it as-is if you don't have strong feelings about it.

lib/avromatic.rb
jcmfernandes: Not really, it was my OCD kicking in, as I have some fun stories to tell about `Array()` and it calling `#to_ary` first. But I can revert it 😅 .

jcmfernandes:
Alright, all done! 😉

lib/avromatic/model/types/map_type.rb
jturkel: I don't have strong feelings about it. Avro [maps](https://avro.apache.org/docs/current/spec.html#Maps) always have string keys so `key_type.referenced_models` should always return the empty array anyway.

jcmfernandes: Oh interesting! As a non-native speaker, it's my understanding that the spec doesn't force keys to be strings, but google says otherwise. https://github.com/apache/avro/pull/17/files even shows "non-string" maps being encoded as arrays to work around this.

I'm going to include this info in the PR.

PR #124, opened 2020-06-29
Compatibility with Avro v1.10.x

tjwp:
Added compatibility with Avro v1.10.x.

avro_turf v1.1.0 has been released with compatibility with Avro v1.10.x.

The Ruby versions for testing are updated and Rails < 5.2 is dropped because those versions no longer receive fixes.


tjwp:
> The change looks good but can you update version.rb and CHANGELOG.md too?

Oops - done! Thanks for the review

PR #125, opened 2020-11-20
Memory optimizations

jturkel:
This PR is part of a series of memory optimizations to our Avro related libraries that reduce serialization memory consumption by 62% and deserialization memory consumption 43% for some of our complex models

A few more details on the optimizations:
* `Avromatic::Model::RawSerialization::Encode#avro_hash` was doing two unnecessary hash allocations
* `Avromatic::Model::Attributes#initialize` was allocating an unnecessary array and hash for each defaulted attribute
* Passing `strict` as a positional argument rather than a keyword argument in `Avromatic::Model::Types::*Type#serialize` avoids a hash allocation on a very hot codepath
* `Avromatic::IO::DatumReader#read_data` was allocating an unnecessary hash for each "optional", non-record attribute value

@will89 - you're prime
/cc @kphelps

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # avromatic changelog
 
+## v2.2.5
+- Optimize memory usage when serializing, deserializing and instantiating models.
+
 ## v2.2.4
 - Compatibility with Avro v1.10.x.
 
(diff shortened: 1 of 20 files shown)


will89 (approved):
Looks great to me! 🚢

PR #127, opened 2020-12-05
Serialization memory optimizations

jturkel:
7% of the memory allocated by our Avromatic serialization benchmarks comes from converting symbolized field names (the format used internally by Avromatic) to string field names (the format used by Avro). Fortunately this information is very cacheable at the model class level :) This optimization improves memory allocation but the throughput improvements are in the margin of error.

@will89 - you're prime

### DIFF
--- a/lib/avromatic/model/configurable.rb
+++ b/lib/avromatic/model/configurable.rb
@@ -8,6 +8,17 @@ module Avromatic
     module Configurable
       extend ActiveSupport::Concern
 
+      # Wraps a reference to a field so we can access both the string and symbolized versions of the name
+      # without repeated memory allocations.
+      class FieldReference
+        attr_reader :name, :name_sym
+
+        def initialize(name)
+          @name = -name
+          @name_sym = name.to_sym
+        end
+      end
+
       module ClassMethods
         attr_accessor :config
         delegate :avro_schema, :value_avro_schema, :key_avro_schema, to: :config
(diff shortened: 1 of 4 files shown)


tjwp:
If you want to require at least Avro v1.10.0, then you can use symbols for the keys in record datums: https://github.com/apache/avro/pull/819/files

jturkel:
@tjwp - Native Avro handling of symbol keys would be very convenient! Unfortunately it looks like this line in [Avro::SchemaValidator.validate_recursive](https://github.com/apache/avro/blob/master/lang/ruby/lib/avro/schema_validator.rb#L97) assumes the datum uses string keys so hashes with symbol keys fail validation. I can put up an Avro PR in the next few weeks to fix that issue but I think we'll want this optimization in the meantime.

tjwp:
> Unfortunately it looks like this line in [Avro::SchemaValidator.validate_recursive](https://github.com/apache/avro/blob/master/lang/ruby/lib/avro/schema_validator.rb#L97) assumes the datum uses string keys so hashes with symbol keys fail validation. I can put up an Avro PR in the next few weeks to fix that issue but I think we'll want this optimization in the meantime.

Ah, thats disappointing. If you do have time to fix it, that would be great!

jturkel:
I opened https://github.com/apache/avro/pull/1036 but I'm assuming that won't get released for awhile so we should move forward with this Avromatic change.

lib/avromatic/model/configurable.rb
```diff
@@ -8,6 +8,17 @@ module Model
     module Configurable
       extend ActiveSupport::Concern
 
+      # Wraps a reference to a field so we can access both the string and symbolized versions of the name
+      # without repeated memory allocations.
+      class FieldReference
+        attr_reader :name, :name_sym
+
+        def initialize(name)
+          @name = -name
```
will89, line 17: Should we freeze this string after dup'ing?

will89 (approved):
Change looks good, just one minor comment.

lib/avromatic/model/configurable.rb
jturkel: It's a bit cryptic but the urinary `-` operator does return a frozen string: https://ruby-doc.org/core-2.7.2/String.html#method-i-2D-40

PR #129, opened 2020-12-09
Simplify/optimize nested model encoding/decoding

jturkel:
This change simplifies the interactions between Avromatic models and lower level Avro serialization/deserialization resulting in:

* Simplified control flow making the code easier to understand
* An Avro serialization/deserialization layer that will be easier to swap out for a native implementation
* Union serialization/deserialization optimizations on non-record union member types that previously only applied to record union member types
* Improved serialization performance for deeply nested schemas - benchmarks show 35% less memory usage and 1.6x throughput increase
* Extensions to `Avro::IO::DatumReader` that duplicate less code from the base classes resulting in less long term maintenance

Now some more details on the changes...

# Serialization

Previously nested models were serialized by Avromatic as a `Hash` with a pointer to the Avromatic model instance (`Avromatic::IO::ENCODING_PROVIDER`) and a union member index (`Avromatic::IO::UNION_MEMBER_INDEX`) if that model appeared in a union field. This enabled the `Avromatic::IO::DatumWriter` to avoid recomputing the union member index since it was already known by Avromatic. Unfortunately the `Avromatic::IO::DatumWriter` callback to the model's `avro_raw_value` method to recursively serialize nested models resulted in extra `StringIO` and `Hash` allocations and calls to `Avro::SchemaValidator.validate!`  which adds lots of overhead for highly nested schemas.

The serialization change is twofold:

1. Recursively convert the Avromatic model into an attributes hash in `Avromatic::Model::Types::RecordType#serialize` before calling the `Avromatic::IO::DatumWriter`.
1. Wrap union datums with `Avromatic::IO::UnionDatum` that includes the member index in `Avromatic::Model::Types::UnionType#serialize` to avoid recomputing the member index in `Avromatic::IO::DatumWriter`.

# Deserialization

Previously the `Avromatic::IO::DatumReader` included a union member's index when deserializing records in a union so Avromatic could optimize the creation of the corresponding Avromatic model instance. With this change the `Avromatic::IO::DatumReader` now wraps the union value in a `Avromatic::IO::UnionDatum` that can be leveraged by `Avromatic::Model::Types::UnionType#serialize`. This also makes it easier for `Avromatic::IO::DatumReader#read_data` to delegate to the super class for everything except the union case we need to handle specially.

@will89 - you're prime
/cc @tjwp @askreet


lib/avromatic/io/datum_reader.rb
```diff
@@ -1,86 +1,37 @@
 # frozen_string_literal: true
 
-# rubocop:disable Style/WhenThen
```
will89, line 3: Can you remove the corresponding enable at the bottom of the file? `rubocop:enable Style/WhenThen`

lib/avromatic/io/datum_reader.rb
```diff
@@ -1,86 +1,37 @@
...
-          raise Avro::IO::SchemaMatchException.new(writers_schema, readers_schema)
-        end
-
+      def read_data(writers_schema, readers_schema, decoder)
         # schema resolution: reader's schema is a union, writer's schema is not
-        if writers_schema.type_sym != :union && readers_schema.type_sym == :union
-          rs_index = readers_schema.schemas.find_index do |s|
-            self.class.match_schemas(writers_schema, s)
-          end
+        return super unless writers_schema.type_sym != :union && readers_schema.type_sym == :union
```
will89, line 12: For my own understanding, this is the change that references reducing the maintenance burden right?

will89 (approved):
Minor suggestions and questions, but this looks great!

lib/avromatic/io/datum_reader.rb
jturkel: Weird. This was removed with a `rubocop -a` but I guess that didn't remove the corresponding `rubocop:enable`.

lib/avromatic/io/datum_reader.rb
jturkel: Yup. This change + removing the override of `read_record` (which was missing https://github.com/apache/avro/pull/636 from upstream).

lib/avromatic/io/union_datum.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      def ==(other)
+        other.is_a?(Avromatic::IO::UnionDatum) &&
+          member_index == other.member_index &&
+          datum == other.datum
+      end
+      alias_method :eql?, :==
+
+      def hash
```
tjwp, line 20: Where is this used?

jturkel: It's not currently used but it makes this class more useful as a true value object. I guess Joshua Bloch's Effective Java book burned it into my head the you should implement the hash code method whenever you implement equals :) I'll add  a unit test to cover the method.

PR #131, opened 2020-12-19
Optimize serialization validation

jturkel:
This PR optimizes model validation during serialization by avoiding a separate validation traversal of the model tree. Product created event instantiation + serialization benchmarks for wide products (198 properties, 75 locales, 3898 localized pval assignments) show a 1.5x throughput increase and 19% less memory allocation.

Most of the complexity of this change comes from ensuring that error messages include the full nested path to any problematic attributes. The code is  intentionally lazy with any error data structures to avoid any extra work for the happy path of a valid model.

I'm considering dropping the explicit `valid?`/`invalid?` methods in Avromatic 3.0 which will remove some of the duplicate validation code this change introduces. Unlike ActiveRecord models, Avromatic models should only be invalid due to a programming error since these Avro "models" are used to represent events that have already happened rather than end user mutable state with complex business validation rules. I'll leave that discussion for another day though :)

@will89 - you're prime
/cc @askreet @cjf2xn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,7 @@
 # avromatic changelog
 
-## v2.3.1
-- Optimize immutable model validation.
-- Don't cache immutable model validation results or serialized Avro attributes if a model has mutable children.
+## v2.4.0
+- Optimize model validation which is now always done during serialization.
 
 ## v2.3.0
 - Add support for Rails 6.1.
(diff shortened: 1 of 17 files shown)


jturkel:
Please hold off reviewing this for a bit. I want to explore an alternate approach that works better with the native serialization work I've been exploring.

jturkel:
Closing in favor of #133

PR #133, opened 2021-01-04
Optimize Serialization Validation (Take 2)

jturkel:
This is an alternative take on #131 that delivers similar performance benefits (1.5x throughput increase and 19% less memory allocation for wide product created event instantiation + serialization benchmarks) by avoiding a separate validation pass over valid model trees. The big difference in the implementation vs #131 is that we just call `avro_validate!` to generate a `Avromatic::Model::ValidationError` with all validation errors as the error propagates up the stack (including from the serialization of nested models). The benefits of this approach over the previous approach are:

1. It's a smaller code change because there's no rewriting of validation error attribute paths.
1. It's easier to implement alternative serialization layers (e.g. a native one) since they can rely on an `avro_validate!` hook to generate an error with complete validation details if they detect a missing required attribute.

@will89 - you're prime
/cc @askreet @cjf2xn


will89 (approved):
Looks good to me!

PR #134, opened 2021-01-05
C++ Encoding/Decoding POC

jturkel:
This is a POC of using the [Avro C++](https://avro.apache.org/docs/current/api/cpp/html/index.html) libraries to optimize Avromatic model encoding/decoding. 

# Benchmark Results

All benchmarks results are relative to the Ruby Avro library version [89c03cfe](https://github.com/apache/avro/tree/89c03cfe5fe8a0ad1aacf910e9b94c57a41ea67e) which includes https://github.com/apache/avro/pull/1014.

## Wide Products

Properties: 198
Locales: 75
Localized Property Values: 6,118

### Encoding
```
                ruby      2.354  (± 0.0%) i/s -     69.000  in  30.299116s
              native      3.773  (±26.5%) i/s -    111.000  in  30.229163s

Comparison:
              native:        3.8 i/s
                ruby:        2.4 i/s - 1.60x  slower
```
### Decoding
```
                ruby      4.281  (± 0.0%) i/s -    128.000  in  30.205893s
              native      6.220  (±16.1%) i/s -    186.000  in  30.109659s

Comparison:
              native:        6.2 i/s
                ruby:        4.3 i/s - 1.45x  slower
```
## Standard Products

Properties: 78
Locales: 4
Localized Property Values: 138

### Encoding
```
                ruby    110.135  (±11.8%) i/s -      3.260k in  30.053549s
              native    166.308  (± 8.4%) i/s -      4.960k in  30.087922s

Comparison:
              native:      166.3 i/s
                ruby:      110.1 i/s - 1.51x  slower
```
### Decoding
```
                ruby    182.973  (± 9.8%) i/s -      5.436k in  30.072907s
              native    227.370  (±10.1%) i/s -      6.732k in  30.010797s

Comparison:
              native:      227.4 i/s
                ruby:      183.0 i/s - 1.24x  slower
```
# Language Choice
So why C++? A few reasons:

1. It's fast
2. It has better memory management facilities than C
3. It has an official Avro library
4. It can leverage the MRI C Extension APIs without any additional third party libraries/frameworks

# Implementation
The core of the encoder implementation is in `ext/avromatic/encoder.cpp`. It walks the Ruby Avromatic model tree converting Ruby objects to [avro::GenericDatum](https://avro.apache.org/docs/current/api/cpp/html/classavro_1_1GenericDatum.html) instances. This datum is then serialized using the Avro C++ API. The core of the decoder implementation is in `ext/avromatic/decoder.cpp`. It uses the Avro C++ API to deserialize the Avro binary data into a tree of [avro::GenericDatum](https://avro.apache.org/docs/current/api/cpp/html/classavro_1_1GenericDatum.html) instances. It then walks this tree of datums creating corresponding Ruby objects. There's some complexity due to how Avromatic handles "optional" fields (i.e. union fields whose first member type is null) but otherwise the conversion between Ruby objects and Avro C++ GenericDatums is pretty straightforward.

A word of warning: I haven't written C++ code in about 18 years so I'm more than a little rusty. The mix of snakecase Ruby C APIs and camelcase Avro/STL APIs is also an eyesore.
# Next Steps
The first things we need to figure out are:

1. Do the performance gains justify the complexity of a native extension?
2. Are we happy using C++ vs. other possible languages (e.g. C or Rust)?

If we decide to move forward with this, then we'll have to figure out the following:
* Packaging - Should the native layer be optional to support JRuby or other environments that can't compile the native extension? Where should we get the Avro C++ libraries from?
* A cleaner separation between the Avromatic model layer and the Avromatic serialization layer over the current conditional hacks.
* Avro validation error handling
* Encoding non-primitive custom types
* Figure out C++ coding conventions e.g. snakecase vs. camelcase, indents, etc.
* Proper testing including things like non-ASCII strings and numeric boundary cases

@will89 @cjf2xn @askreet @kphelps - I'm curious for your feedback on this. No rush since I probably won't get back around to working on it for a few weeks.
/cc @salsify/kafka-developers

### DIFF
--- a/ext/avromatic/decoder.cpp
+++ b/ext/avromatic/decoder.cpp
@@ -1,0 +1,192 @@
+#include <encoder.hpp>
+#include <common.hpp>
+
+#include <iostream>
+
+#include <ruby/encoding.h>
+
+#include <avro/Decoder.hh>
+#include <avro/Generic.hh>
+#include <avro/GenericDatum.hh>
+#include <avro/Specific.hh>
+
+static VALUE rb_avromatic_native_module;
+static VALUE rb_union_datum_class;
+
+VALUE utf8_string_to_ruby(const std::string& value) {
+  static rb_encoding* encoding = rb_enc_find("UTF-8");
+  return rb_enc_str_new(value.c_str(), value.size(), encoding);
+}
+
+VALUE deserialize_logical_date(int value) {
+  VALUE rb_epoch_start = rb_const_get(rb_avromatic_native_module, rb_intern("EPOCH_START"));
+  return rb_funcall(rb_epoch_start, rb_intern("+"), 1, INT2NUM(value));
+}
+
+VALUE deserialize_logical_timestamp_millis(long value) {
+  long seconds = value / 1000;
+  long milliseconds = value % 1000;
+  VALUE time_class = rb_const_get(rb_cObject, rb_intern("Time"));
+  VALUE rb_value = rb_funcall(time_class, rb_intern("at"), 2, LONG2NUM(seconds), LONG2NUM(milliseconds * 1000));
+  return rb_funcall(rb_value, rb_intern("utc"), 0);
+}
+
+VALUE deserialize_logical_timestamp_micros(long value) {
+  long seconds = value / 1000000;
+  long microseconds = value % 1000000;
+  VALUE time_class = rb_const_get(rb_cObject, rb_intern("Time"));
+  VALUE rb_value = rb_funcall(time_class, rb_intern("at"), 2, LONG2NUM(seconds), LONG2NUM(microseconds));
+  return rb_funcall(rb_value, rb_intern("utc"), 0);
+}
+
+bool is_optional_field(const avro::NodePtr& schema) {
+  return schema->type() == avro::AVRO_UNION &&
+    schema->leafAt(0)->type() == avro::AVRO_NULL &&
+    schema->leaves() == 2;
+}
+
+// Unfortunately we need to track the schema for unions due to https://issues.apache.org/jira/browse/AVRO-2597
+VALUE datum_to_ruby_value(const avro::NodePtr& schema, const avro::GenericDatum& datum, bool strict) {
+  VALUE rb_value;
+  switch (datum.type()) {
+    case avro::AVRO_STRING:
+    {
+      const std::string& field_value = datum.value<std::string>();
+      rb_value = utf8_string_to_ruby(field_value);
+      break;
+    }
+    case avro::AVRO_BYTES:
+    {
+      const std::vector<uint8_t>& bytes = datum.value<std::vector<uint8_t> >();
+      rb_value = rb_str_new((const char*)&bytes[0], bytes.size());
+      break;
+    }
+    case avro::AVRO_INT:
+    {
+      if (datum.logicalType().type() == avro::LogicalType::DATE) {
+        rb_value = deserialize_logical_date(datum.value<int>());
+      } else {
+        rb_value = INT2NUM(datum.value<int>());
+      }
+      break;
+    }
+    case avro::AVRO_LONG:
+    {
+      if (datum.logicalType().type() == avro::LogicalType::TIMESTAMP_MILLIS) {
+        rb_value = deserialize_logical_timestamp_millis(datum.value<long>());
+      } else if (datum.logicalType().type() == avro::LogicalType::TIMESTAMP_MICROS) {
+        rb_value = deserialize_logical_timestamp_micros(datum.value<long>());
+      } else {
+        rb_value = LONG2NUM(datum.value<long>());
+      }
+      break;
+    }
+    case avro::AVRO_FLOAT:
+    {
+      rb_value = DBL2NUM(datum.value<float>());
+      break;
+    }
+    case avro::AVRO_DOUBLE:
+    {
+      rb_value = DBL2NUM(datum.value<double>());
+      break;
+    }
+    case avro::AVRO_BOOL:
+    {
+      rb_value = datum.value<bool>() ? Qtrue : Qfalse;
+      break;
+    }
+    case avro::AVRO_NULL:
+    {
+      rb_value = Qnil;
+      break;
+    }
+    case avro::AVRO_RECORD:
+    {
+      rb_value = rb_hash_new();
+      const avro::GenericRecord& record_value = datum.value<avro::GenericRecord>();
+      const int num_fields = (int)record_value.fieldCount();
+      for (int i=0; i<num_fields; ++i) {
+        const std::string& field_name = record_value.schema()->nameAt(i);
+        VALUE rb_field_name = ID2SYM(rb_intern(field_name.c_str()));
+        const avro::GenericDatum& field_datum = record_value.fieldAt(i);
+        rb_hash_aset(rb_value, rb_field_name, datum_to_ruby_value(record_value.schema()->leafAt(i), field_datum, strict));
+      }
+      break;
+    }
+    case avro::AVRO_ENUM:
+    {
+      const std::string& symbol = datum.value<avro::GenericEnum>().symbol();
+      rb_value = ID2SYM(rb_intern(symbol.c_str()));
+      break;
+    }
+    case avro::AVRO_ARRAY:
+    {
+      const avro::GenericArray::Value& array_value = datum.value<avro::GenericArray>().value();
+      const long num_elements = array_value.size();
+      rb_value = rb_ary_new_capa(num_elements);
+      for (long i=0; i<num_elements; ++i) {
+        rb_ary_store(rb_value, i, datum_to_ruby_value(schema->leafAt(0), array_value[i], strict));
+      }
+      break;
+    }
+    case avro::AVRO_MAP:
+    {
+      rb_value = rb_hash_new();
+      const avro::GenericMap::Value& map_entries = datum.value<avro::GenericMap>().value();
+      for (auto iter = map_entries.cbegin(); iter != map_entries.cend(); ++iter) {
+        VALUE rb_map_key = utf8_string_to_ruby(iter->first);
+        VALUE rb_map_value = datum_to_ruby_value(schema->leafAt(1), iter->second, strict);
+        rb_hash_aset(rb_value, rb_map_key, rb_map_value);
+      }
+     break;
+    }
+    case avro::AVRO_FIXED:
+    {
+      const avro::GenericFixed& fixedValue = datum.value<avro::GenericFixed>();
+      const std::vector<uint8_t>& bytes = fixedValue.value();
+      rb_value = rb_str_new((const char*)&bytes[0], bytes.size());
+      break;
+    }
+    default:
+      rb_raise(rb_eRuntimeError, "Unexpected field type %s", avro::toString(datum.type()).c_str());
+      // Make the compiler happy
+      rb_value = Qnil;
+  }
+
+  if (!strict && schema->type() == avro::AVRO_UNION && !is_optional_field(schema) && datum.type() != avro::AVRO_NULL) {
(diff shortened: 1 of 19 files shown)


ext/avromatic/decoder.cpp
```diff
@@ -0,0 +1,192 @@
...
+    case avro::AVRO_FIXED:
+    {
+      const avro::GenericFixed& fixedValue = datum.value<avro::GenericFixed>();
+      const std::vector<uint8_t>& bytes = fixedValue.value();
+      rb_value = rb_str_new((const char*)&bytes[0], bytes.size());
+      break;
+    }
+    default:
+      rb_raise(rb_eRuntimeError, "Unexpected field type %s", avro::toString(datum.type()).c_str());
+      // Make the compiler happy
```
askreet, line 153: My understanding of the Ruby API is that this isn't just making the compiler happy, `rb_raise` does _not_ stop execution of the extension. Instead, it flags the VM to enter exception handling mode with this exception _when control returns to the VM_. Thus, this value is returned to Ruby, and, I think, bound to the lhs of the statement.

ext/avromatic/encoder.cpp
```diff
@@ -0,0 +1,267 @@
...
+      avro::GenericMap& mapValue = datum.value<avro::GenericMap>();
+      VALUE rb_avromatic_item_type = rb_funcall(rb_avromatic_type, rb_intern("value_type"), 0);
+      // TODO: Avoid allocating the key array by using rb_hash_foreach
+      VALUE rb_keys = rb_funcall(rb_value, rb_intern("keys"), 0);
+      const long num_keys = rb_array_len(rb_keys);
+      mapValue.value().resize(num_keys, std::pair<std::string, avro::GenericDatum>("", avro::GenericDatum(mapValue.schema()->leafAt(1))));
+      for (long i=0; i<num_keys; ++i) {
+        VALUE rb_key = rb_ary_entry(rb_keys, i);
+        VALUE rb_utf8_key = rb_funcall(rb_key, rb_intern("encode"), 1, rb_id2str(rb_intern("utf-8")));
+        // TODO: Does this work correctly with utf-8 characters?
```
askreet, line 188: I just did some reading on this and, as far as I can tell, the only time a zero byte should appear in UTF-8 is when it's encoding a literal NUL. For all multibyte characters the values start at 127.

I also found that there is `ruby/encoding.h` lib which seems to have ways to set encoding on rb String types directly from C, `rb_encoding_find_index("UTF-8")`, and subsequently `rb_encoding_associate_index(index, rb_key)`. The docs aren't crystal clear though so I'm not sure it does what you want here. I tried using it like so:

```cpp
        VALUE rb_key = rb_ary_entry(rb_keys, i);
        auto utf8 = rb_enc_find_index("UTF-8");
        rb_enc_associate_index(utf8, rb_key);
```

And then the Ruby code complained that I can't modify a frozen Integer, which, I found baffling. I gave up at that point. :-)

ext/avromatic/encoder.cpp
```diff
@@ -0,0 +1,267 @@
...
+        mapValue.value()[i].first = StringValueCStr(rb_utf8_key);
```
askreet, line 189: In case I'm mislead about UTF-8 not having a NUL byte during encoding, I think the safer option here is an explicit:

```cpp
        mapValue.value()[i].first = std::string(StringValuePtr(rb_utf8_key), RSTRING_LEN(rb_utf8_key));
```

This might have the micro-optimization of not having to traverse the `const char*` to find the length when building the `std::string` structure and instead just chunk the memory over. 🤷🏼 

(Tested on my machine.)

ext/avromatic/encoder.cpp
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  std::unique_ptr<avro::OutputStream> outputStream = avro::memoryOutputStream();
+  avro::EncoderPtr encoder = avro::validatingEncoder(*schema, avro::binaryEncoder());
+  encoder->init(*outputStream);
+  try {
+    avro::encode(*encoder, datum);
+  } catch (const avro::Exception &e) {
+    // This should never happen since the model should be valid at this point but just in case
+    rb_raise(rb_avromatic_validation_error_class, "%s", e.what());
```
askreet, line 249: IIUC we probably want to return a nil here, rather than trying to execute line 253.

askreet (commented):
This is awesome :+1:.

A chunk of my feedback is just C++17 patterns we could use, totally optional.

askreet:
I should add that I really only looked at the C++ portion of this PR. I'm not familiar enough with the use cases around the Ruby API to feel comfortable weighing in.

kphelps (commented):
Code looks pretty good to me overall. I think another more modern C++ style point is to embrace `auto` for type inference - in general you can use this for all local variables.

> Do the performance gains justify the complexity of a native extension?

Seems like it. A ~50% improvement is pretty awesome and its a pretty small footprint. I'm assuming there is more room for meaningful optimization to continue iterating on this, though. 

> Are we happy using C++ vs. other possible languages (e.g. C or Rust)?

I think C++ is a good choice here - The avro C++ library actually looks quite nice to work with. If this gets significantly bigger it might be a good idea to look into something like https://github.com/jasonroelofs/rice to have a C++ interface to the ruby VM for added type safety / ease of use.

> Figure out C++ coding conventions e.g. snakecase vs. camelcase, indents, etc.

Pulling in the `clang` tools (`clang-format` and `clang-tidy` in particular) would be helpful for this.

ext/avromatic/decoder.cpp
```diff
@@ -0,0 +1,192 @@
...
+      for (long i=0; i<num_elements; ++i) {
+        rb_ary_store(rb_value, i, datum_to_ruby_value(schema->leafAt(0), array_value[i], strict));
+      }
+      break;
+    }
+    case avro::AVRO_MAP:
+    {
+      rb_value = rb_hash_new();
+      const avro::GenericMap::Value& map_entries = datum.value<avro::GenericMap>().value();
+      for (auto iter = map_entries.cbegin(); iter != map_entries.cend(); ++iter) {
```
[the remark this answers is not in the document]
jturkel: Very cool. C++ has come a long way since 2003!

jturkel:
Thanks for the feedback @askreet and @kphelps! My antiquated C++ aside, it sounds like you guys are onboard with a C++ extension to optimize encoding/decoding performance. I'll fold in the feedback you've given so far and then start working through the TODO list in the PR description. Ping me if you have the bandwidth and desire to pickup any of the remaining TODOs.

ext/avromatic/decoder.cpp
jturkel: This turns out to be quite a thorny issue! `rb_raise` does in fact stop execution of the C extension immediately using `longjmp` (with a corresponding `setjmp` before the extension is invoked) which is bad since it means no C++ destructors are invoked as you pop back up the stack. I think the way to fix this is to use C++ exceptions throughout most of the extension and then have very carefully crafted top level "gateway" functions that don't rely on any stack allocated object destructors and invoke `rb_raise` e.g.

```c
VALUE encode_model_with_error_handling(VALUE self, VALUE rb_model, VALUE rb_is_value_schema) {
  VALUE result = encode_model(self, rb_model, rb_is_value_schema);
  if (rb_obj_is_kind_of(result, rb_eException) == Qtrue) {
    rb_exc_raise(result);
  }
  return result;
}
```
This highlights another problem though. Any time the C extension calls back into Ruby with `rb_funcall` a similar `longjmp` will happen if that Ruby method throws an exception leading to memory leaks. The fix is use `rb_protect` so you can convert the Ruby exception to a C++ exception  that can then be handled further up the stack:

```
int state;
VALUE result = rb_protect(dangerous_func, dangerous_arg, &state);

if (state) {
  // an exception occurred
  VALUE err = rb_errinfo();
  throw RubyException(err);
}
else {
  // no exception occurred
  return result;
}
```
Gross. Fortunately the extension only makes two calls back into Ruby that could realistically throw exceptions so the problem isn't too widespread. 

All of this does make me think I should take another look at [rice](https://github.com/jasonroelofs/rice) though. I had initially ruled that library out since it I was worried the abstraction layer wouldn't really encapsulate the underlying Ruby APIs resulting in developers needing to understand both the Ruby APIs and the Rice APIs but maybe the error handling justifies this risk.

askreet: Wow, what a mess. This makes it so easy to leak memory! Having the error handling convert C++ exceptions to Ruby exceptions as few times as possible sounds promising, as does using `rice`. I wonder how many native extensions do this wrong and just leak bits of memory here and there.

skarger:
> Do the performance gains justify the complexity of a native extension?

This question makes me wonder if Ruby's MJIT could produce a similar improvement.

However it sounds like MJIT will not be suitable for Rails until at least Ruby 3.1:
https://medium.com/@k0kubun/ruby-3-0-jit-and-beyond-4d9404ce33c

Since in practice the encoding/decoding needs to happen within a Rails process, that seems like a blocker. But could the benchmark run as a standalone Ruby 3.0 program that only depends on Avromatic?

jturkel:
@skarger - I agree MJIT is intriguing but pretty far from being production ready for Rails apps. It does reenforce my belief that we should make the Avromatic persistence layer pluggable so we can hopefully move back to a pure Ruby implementation some day. Writing C++ again did bring back fond memories but I'd rather keep it relegated to my past :)

I had an earlier attempt at pushing parts of Ruby Avro down into C over in https://github.com/salsify/fast-avro-poc but the results weren't very impressive. Effectively I was manually JITing the Ruby code. I think the biggest problem was there was too much back and forth between the C layer and the Ruby layer which prevented lots of compiler optimization. Nonetheless that repo does have some useful benchmarks of the lower level Avro APIs (i.e. the benchmarks don't instantiate Avromatic models). I did a quick test w/ Ruby 3.0.0 with and without JIT and JIT made things slower:

```
Ruby 2.7.2
       Avro Encoding    101.041  (± 3.0%) i/s -    510.000  in   5.051473s
Avromatic Optimized Encoding
                        398.744  (± 2.3%) i/s -      2.028k in   5.088403s
       Avro Decoding    634.422  (± 1.6%) i/s -      3.213k in   5.065636s
Avromatic Optimized Decoding
                        586.952  (± 1.9%) i/s -      2.950k in   5.027883s

Ruby 3.0.0
       Avro Encoding    103.503  (± 5.8%) i/s -    520.000  in   5.047783s
Avromatic Optimized Encoding
                        438.557  (± 3.0%) i/s -      2.226k in   5.080289s
       Avro Decoding    627.787  (± 1.3%) i/s -      3.162k in   5.037532s
Avromatic Optimized Decoding
                        586.689  (± 2.0%) i/s -      2.958k in   5.043956s

Ruby 3.0.0 w/ JIT (RUBYOPT="--jit")
       Avro Encoding     94.714  (± 7.4%) i/s -    477.000  in   5.067347s
Avromatic Optimized Encoding
                        397.405  (± 8.8%) i/s -      1.974k in   5.016216s
       Avro Decoding    580.139  (±11.9%) i/s -      2.860k in   5.013489s
Avromatic Optimized Decoding
                        576.228  (± 4.9%) i/s -      2.907k in   5.057432s
```
I didn't play around with the various JIT options so maybe that could speed things up.

skarger:
> I did a quick test w/ Ruby 3.0.0 with and without JIT and JIT made things slower.

That's too bad. On the bright side, apparently encoding with non-JITed Ruby 3.0 is faster than 2.7 for both optimized and not.

jturkel:
I reran the standard product benchmarks for Ruby 2.7 vs. Ruby 3.2 w/ YJIT and found that Ruby 3.2 /w YJIT was 2.0x faster decoding and 1.6x faster encoding. Unfortunately I'm running into some issues compiling the native Avromatic extension on my new laptop so I can't include that in the performance comparison but the improvement of Ruby 3.2 w/ YJIT is compelling enough that I'm not planning on pursuing a native extension further.

PR #136, opened 2021-04-21
Fix boolean unions when the messages has a True value

fgarces:
When using unions with booleans, messages fail to decode if they contain a true value. 

@cgrdavies  you are prime

cc/ @tjwp since @jturkel is on vacation


.gitignore
```diff
@@ -9,3 +9,6 @@
 /tmp/
 /gemfiles/*.lock
 /log/*
+
+*.iml
+.idea
```
cgrdavies, line 14: we might want to keep editor-specific stuff out of here

cgrdavies (approved):
Looks good to me!

tjwp (approved):
One minor comment, but makes sense to me. Let me know if you'd like me to push a new 3.0 release with this fix.

fgarces:
@tjwp I just merged and it would be great if you could push the release.

PR #137, opened 2021-06-07
Detect already registered nested models

jturkel:
This PR fixes a regression introduced by #119 that caused Avromatic to not detect models that conflict with already generated models due to improper eager loading configuration. This bug lead to hard to troubleshoot downstream errors like this when instantiating model instances:

```
Avromatic::Model::CoercionError: Value for TargetSchemas::Events::ApplicableScope#field_conditions could not be coerced to a array[FieldCondition]. Provided argument: [#<TargetSchemas::Events::FieldCondition field: #<Core::Events::Reference id: SalsifyUuid('s-8073254e-9d10-487f-8749-a44d04519f1b'), type: "target_schema_fields", external_id: "Field 000000000 ID">, values: ["foo", "bar"], operator: nil, quantifier: nil>]
```

With this bug fix we now get a more actionable error message at model definition time:

```
RuntimeError: Attempted to replace existing Avromatic model FieldCondition with new model TargetSchemas::Events::FieldCondition as 'target_schemas.applicable_scopes.field_condition'. Perhaps 'TargetSchemas::Events::FieldCondition' needs to be eager loaded via the Avromatic eager_load_models setting?
```

@kphelps - you're prime
/cc @salsify/kafka-developers @tjwp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # avromatic changelog
 
+## 3.0.1
+- Raise an error when registering a nested model that has already been auto-generated.
+  This avoids hard to troubleshoot coercion errors when instantiating models and fixes 
+  a regression introduced in Avromatic 2.2.2.
+
 ## 3.0.0
 - Drop support for Ruby 2.4.
 - Add support for Ruby 3.0.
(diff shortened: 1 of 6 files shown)


kphelps (approved):
:ship: Nice find.

PR #138, opened 2021-07-08
Ensure schema registry caches are cleared between tests

jturkel:
This PR updates the RSpec integration to clear schema registry caches are cleared between tests which ensures consistent ID mappings with the `AvroSchemaRegistry::FakeServer`. As part of this fix I added an `Avromatic.build_schema_registry!` method for consistency with the existing `Avromatic.build_messaging!` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # avromatic changelog
 
+## 3.0.2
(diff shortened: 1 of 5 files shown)


cjf2xn (approved):
🎉

PR #140, opened 2021-12-27
Support Avro 1.11 and Rails 7.0, drop Ruby 2.6 and Avro 1.9 support

tjwp:
For compatibility with Ruby 2.7, I updated to the latest `salsify_rubocop` (v1.1.0) and fixed offenses. Those changes are in a separate commit.

I've omitted testing against Ruby 3.1.0 for now because there is an incompatibility between Rails 7.0.0 and Ruby 3.1.0: https://github.com/rails/rails/pull/43970


avromatic.gemspec
```diff
@@ -20,11 +20,13 @@ Gem::Specification.new do |spec|
...
+  spec.metadata['rubygems_mfa_required'] = 'true'
 
-  spec.add_runtime_dependency 'activemodel', '>= 5.2', '< 6.2'
-  spec.add_runtime_dependency 'activesupport', '>= 5.2', '< 6.2'
-  spec.add_runtime_dependency 'avro', '>= 1.9.0', '< 1.11'
+  spec.required_ruby_version = '>= 2.7'
+
+  spec.add_runtime_dependency 'activemodel', '>= 5.2', '< 7.1'
+  spec.add_runtime_dependency 'activesupport', '>= 5.2', '< 7.1'
+  spec.add_runtime_dependency 'avro', '>= 1.9.0', '< 1.12'
```
jturkel, line 29: How would you feel about dropping Avro 1.9 support since we're already doing a major version bump?

jturkel (commented):
One question but otherwise looks great.

avromatic.gemspec
tjwp: No objection since it is already almost two years old. I've added a commit that does that. I can squash the commits before merging.

At the same time, I added testing against Ruby 3.1 since there is now a compatible Rails 7.0 release.

I also excluded the `gemfiles/` directory for Rubocop so that flagging offenses and auto-correcting those generated files.

PR #142, opened 2022-02-02
Add support for specifying a subject for the avro schema when building an Avromatic model

will89:
This is an attempt to address https://github.com/salsify/avromatic/issues/141. It seems like some tools like ksqlDB have schema naming conventions that are not allowed within the avro naming conventions. ksqlDB expects all schemas to be ending in a `-value` in the schema registry, https://docs.ksqldb.io/en/latest/operate-and-deploy/schema-registry-integration/.

prime: @jturkel

### DIFF
--- a/lib/avromatic/model/configuration.rb
+++ b/lib/avromatic/model/configuration.rb
@@ -7,7 +7,7 @@ module Avromatic
     class Configuration
 
       attr_reader :avro_schema, :key_avro_schema, :nested_models, :mutable,
-                  :allow_optional_key_fields
+                  :allow_optional_key_fields, :avro_schema_subject, :key_avro_schema_subject
       alias_method :mutable?, :mutable
       delegate :schema_store, to: Avromatic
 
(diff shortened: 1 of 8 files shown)


lib/avromatic/model/configuration.rb
```diff
@@ -17,24 +17,30 @@ class Configuration
...
+      # @option options [String, Symbol] :value_schema_subject
       # @option options [Avro::Schema] :key_schema
       # @option options [String, Symbol] :key_schema_name
+      # @option options [String, Symbol] :key_schema_subject
       # @option options [Avromatic::ModelRegistry] :nested_models
       # @option options [Boolean] :mutable, default false
       # @option options [Boolean] :allow_optional_key_fields, default false
       def initialize(**options)
         @avro_schema = find_avro_schema(**options)
+        @avro_schema_subject = options[:schema_subject] || options[:value_schema_subject]
```
jturkel, line 32: Would it make sense to default the subject to the schema's full name here to avoid defaulting in a few other places that could get out of sync?

spec/avromatic/model/builder_spec.rb
```diff
@@ -141,6 +141,21 @@
       end
 
       it_behaves_like "a generated model"
+
+      context "with a specified schema subject" do
+        let(:schema_subject) { 'test.primitive_types-value' }
```
jturkel, line 146: nit: Some of this test data might be easier to understand if we use the `-subject` suffix rather than `-value` (especially for the key schemas).

jturkel (approved):
A few minor comments but looks good!

lib/avromatic/model/configuration.rb
will89: Were you thinking of something like `@avro_schema_subject = options[:schema_subject] || options[:value_schema_subject] || avro_schema.fullname`?

This seems to be problematic if the schema is a `Avro::Schema::UnionSchema` since it does not have a `fullname`.
I got this failure `rspec ./spec/avromatic/model/builder_spec.rb:257 # Avromatic::Model::Builder model generation top-level union raises an error`.

Also, I think this only helps with the defaulting in the method `Registration#register_schema`. Since the messaging no longer has the model context. I don't know if this defaulting is worthwhile.

IanWhitney:
Thanks!

PR #151, opened 2023-01-27
Add support for decimal logical type

opti:
It adds support for the decimal logical type, so the following Avro schema can be processed correctly:
```json
{
  "type": "bytes",
  "logicalType": "decimal",
  "precision": 4,
  "scale": 2
}
```
However, It's only supported by the Avro v1.11.x. If that's ok, I'm open for suggestions how to better restrict that feature to supported avro version.

https://avro.apache.org/docs/1.11.1/specification/#decimal

### DIFF
--- a/lib/avromatic/model/types/decimal_type.rb
+++ b/lib/avromatic/model/types/decimal_type.rb
@@ -10,8 +10,14 @@ module Avromatic
       class DecimalType < AbstractType
         VALUE_CLASSES = [::BigDecimal].freeze
         INPUT_CLASSES = [::Numeric, ::String].freeze
-        DEFAULT_PRECISION = 3
-        private_constant :DEFAULT_PRECISION
+
+        attr_reader :precision, :scale
+
+        def initialize(precision:, scale: 0)
+          super()
+          @precision = precision
+          @scale = scale
+        end
 
         def value_classes
           VALUE_CLASSES
@@ -22,7 +28,7 @@ module Avromatic
         end
 
         def name
-          'decimal'
+          "decimal(#{precision}, #{scale})"
         end
 
         def coerce(input)
@@ -30,7 +36,7 @@ module Avromatic
           when ::NilClass, ::BigDecimal
             input
           when ::Rational
-            input.to_d(DEFAULT_PRECISION)
+            input.to_d(precision)
           when ::Numeric, ::String
             input.to_d
           else
--- a/lib/avromatic/model/types/decimal_type.rb
+++ b/lib/avromatic/model/types/decimal_type.rb
@@ -9,7 +9,6 @@ module Avromatic
     module Types
       class DecimalType < AbstractType
         VALUE_CLASSES = [::Numeric].freeze
-        INPUT_CLASSES = [::Numeric, ::String].freeze
 
         attr_reader :precision, :scale
 
@@ -23,10 +22,6 @@ module Avromatic
           VALUE_CLASSES
         end
 
-        def input_classes
-          INPUT_CLASSES
-        end
-
         def name
           "decimal(#{precision}, #{scale})"
         end
@@ -37,7 +32,7 @@ module Avromatic
             input
           when ::Rational
             input.to_d(precision)
-          when ::Numeric, ::String
+          when ::Numeric
             input.to_d
           else
             raise ArgumentError.new("Could not coerce '#{input.inspect}' to #{name}")
@@ -45,7 +40,7 @@ module Avromatic
         end
 
         def coercible?(input)
-          input.nil? || input.is_a?(::String) || input.is_a?(::Numeric)
+          input.nil? || input.is_a?(::Numeric)
         end
 
         def coerced?(input)
--- a/lib/avromatic/model/types/decimal_type.rb
+++ b/lib/avromatic/model/types/decimal_type.rb
@@ -30,8 +30,6 @@ module Avromatic
           case input
           when ::NilClass, ::BigDecimal
             input
-          when ::Rational
-            input.to_d(precision)
           when ::Numeric
             input.to_d
           else
--- a/lib/avromatic/model/types/decimal_type.rb
+++ b/lib/avromatic/model/types/decimal_type.rb
@@ -8,7 +8,7 @@ module Avromatic
   module Model
     module Types
       class DecimalType < AbstractType
-        VALUE_CLASSES = [::BigDecimal].freeze
+        VALUE_CLASSES = [::Numeric].freeze
         INPUT_CLASSES = [::Numeric, ::String].freeze
 
         attr_reader :precision, :scale
--- a/lib/avromatic/model/types/decimal_type.rb
+++ b/lib/avromatic/model/types/decimal_type.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'bigdecimal'
+require 'bigdecimal/util'
+require 'avromatic/model/types/abstract_type'
+
+module Avromatic
+  module Model
+    module Types
+      class DecimalType < AbstractType
+        VALUE_CLASSES = [::BigDecimal].freeze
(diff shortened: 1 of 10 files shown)


lib/avromatic/model/types/decimal_type.rb
```diff
@@ -0,0 +1,59 @@
...
+        def name
+          'decimal'
+        end
+
+        def coerce(input)
+          case input
+          when ::NilClass, ::BigDecimal
+            input
+          when ::Rational
+            input.to_d(DEFAULT_PRECISION)
```
jturkel, line 33: Should we be truncating `Float`s too?

lib/avromatic/model/types/decimal_type.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'bigdecimal'
+require 'bigdecimal/util'
```
jturkel, line 4: Why is this required?

lib/avromatic/model/types/decimal_type.rb
```diff
@@ -0,0 +1,59 @@
...
+        def value_classes
+          VALUE_CLASSES
+        end
+
+        def input_classes
+          INPUT_CLASSES
+        end
+
+        def name
+          'decimal'
```
jturkel, line 25: Can we include the `precision` and `scale` in the name?

lib/avromatic/model/types/type_factory.rb
```diff
@@ -26,6 +27,7 @@ module TypeFactory
           'date' => Avromatic::Model::Types::DateType.new,
           'timestamp-micros' => Avromatic::Model::Types::TimestampMicrosType.new,
           'timestamp-millis' => Avromatic::Model::Types::TimestampMillisType.new,
+          'decimal' => Avromatic::Model::Types::DecimalType.new,
```
jturkel, line 30: Could we conditionally add this based on `Avro::VERSION` so that we only support decimals in Avro 1.11.0+?

jturkel (changes requested):
Apologies for the delay reviewing this @opti! It would be awesome to include decimal support in Avromatic.

lib/avromatic/model/types/decimal_type.rb
opti: This provides `to_d` method for other types like `Float`, `Integer`, `String`, etc.

lib/avromatic/model/types/decimal_type.rb
```diff
@@ -0,0 +1,59 @@
...
+
+require 'bigdecimal'
+require 'bigdecimal/util'
+require 'avromatic/model/types/abstract_type'
+
+module Avromatic
+  module Model
+    module Types
+      class DecimalType < AbstractType
+        VALUE_CLASSES = [::BigDecimal].freeze
```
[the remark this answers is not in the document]
opti: My impression was that `VALUE_CLASSES` is for the return type and not the input for the Avro#encode. Would appreciate any details here.

lib/avromatic/model/types/decimal_type.rb
```diff
@@ -0,0 +1,59 @@
...
+        INPUT_CLASSES = [::Numeric, ::String].freeze
+        DEFAULT_PRECISION = 3
```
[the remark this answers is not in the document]
opti: I wasn't sure if Rational needs to be explicitly supported at all. It's Numeric, but Avro ruby library [doesn't do any specific coercion](https://github.com/apache/avro/blob/4e1fefca493029ace961b7ef8889a3722458565a/lang/ruby/lib/avro/logical_types.rb#L122) for it, even though they coerce value with `to_d`, which requires precision to be provided.

I can rework it to use `precision` from the schema.

lib/avromatic/model/types/decimal_type.rb
opti: Wouldn't it be better to delegate that to Avro? https://github.com/apache/avro/blob/4e1fefca493029ace961b7ef8889a3722458565a/lang/ruby/lib/avro/logical_types.rb#L180-L200

lib/avromatic/model/types/decimal_type.rb
```diff
@@ -0,0 +1,59 @@
...
+          'decimal'
+        end
+
+        def coerce(input)
+          case input
+          when ::NilClass, ::BigDecimal
+            input
+          when ::Rational
+            input.to_d(DEFAULT_PRECISION)
+          when ::Numeric, ::String
```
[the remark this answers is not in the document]
opti: BigDecimal supports [coercion for strings](https://docs.ruby-lang.org/en/3.2/String.html#method-i-to_d) out of the box and it seems safe enough.

In comparison to `FloatType` and `IntegerType`, Avro's schema accepts `bytes` as the main type for the `logicalType: "decimal"`, which means String should be accepted.

opti:
@jturkel Thank you for the review. I've made updates for some of your comments and provided feedback for others.

lib/avromatic/model/types/decimal_type.rb
opti: After more consideration it seems that supporting coercion for `Rational` doesn't make a lot of sense here, because eventually it stumbles into the rounding issue anyway. And if I understand correctly the Avro ruby library's approach, they avoid rounding and instead delegate it to the user by raising `Rounding necessary` error when not rounded value provided.

The only benefit of having support for `Rational` coercion is to not raise error when `input.to_d` is called without `precision`, which is [required by specs](https://ruby-doc.org/3.2.1/exts/bigdecimal/Rational.html#method-i-to_d)

opti:
@jturkel any chance for another review?

lib/avromatic/model/types/decimal_type.rb
jturkel: I agree with your assessment. We're safer following the same approach as Avro Ruby and delegating any rounding decisions to the client. Thus no need to support `Rational`.

lib/avromatic/model/types/decimal_type.rb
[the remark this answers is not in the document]
opti: Thanks for the explanation. That makes sense. Updated value classes to `[::Numeric]`.

lib/avromatic/model/types/decimal_type.rb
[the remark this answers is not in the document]
opti: `String` has been removed from supported inputs.

lib/avromatic/model/types/decimal_type.rb
```diff
@@ -0,0 +1,58 @@
...
+
+        def name
+          "decimal(#{precision}, #{scale})"
+        end
+
+        def coerce(input)
+          case input
+          when ::NilClass, ::BigDecimal
+            input
+          when ::Numeric
```
jturkel, line 33: Can we restrict this conditional to just `Float` and `Integer` since we're not supporting `Rational` or `Complex`? The implementation of `coercible?` will also require a similar change.

jturkel (changes requested):
One small thing I noticed during the final review but otherwise looks great!

lib/avromatic/model/types/decimal_type.rb
opti: @jturkel Should the `VALUE_CLASSES` and `INPUT_CLASSES` be limited to `[BigDecimal, Float, Integer]` as well then?

jturkel: `DecimalType#input_classes` should return `[BigDecimal, Float, Integer]` because those are the types that we support as input for decimal attributes in constructors and attribute setters. `DecimalType#value_classes` should return `[BigDecimal]` because decimal attribute values are always coerced to a `BigDecimal` and a decimal attribute getter will always return a `BigDecimal`.

opti: Make sense. Updated.

PR #152, opened 2023-02-08
Updates readme to make it a little clearer when configuring a schema registry

mkrisher:
This PR updates the README to make it a little clearer when configuring a schema registry that you have the option to either provide an instance of a compatible registry object or just specify the registry_url. `build_schema_registry` uses the registry_url to instantiate an instance of a compatible registry class.

Setting schema_registry and a registry_url, followed by a call to build_schema_registry will override whatever was passed in as the value of schema_registry with the instantiation of the ConfluentSchemaRegistry instance configured with the registry_url.

This also updates the example to show that `build_messaging` will call `build_schema_registry`. That both do not need to be called explicitly (at least when providing registry_url).

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Avromatic.configure do |config|
 end
 ```
 
-NOTE: `build_message!` ultimately calls `build_schema_registry!` so you don't have
+NOTE: `build_messaging!` ultimately calls `build_schema_registry!` so you don't have
 to call both.
 
 #### Decoding
--- a/README.md
+++ b/README.md
@@ -99,6 +99,9 @@ Avromatic.configure do |config|
 end
 ```
 
+NOTE: `build_message!` ultimately calls `build_schema_registry!` so you don't have
+to call both.
+
 #### Decoding
 
 * **use_custom_datum_reader**: `Avromatic` includes a modified subclass of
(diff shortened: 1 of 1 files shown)


joshbranham:
Thanks for the PR @mkrisher! Unsure why CircleCI is not running for the fork, reaching out to their support to see if we can figure that out first.

README.md
```diff
@@ -93,7 +94,7 @@ Example using a schema registry:
 Avromatic.configure do |config|
   config.schema_store = AvroTurf::SchemaStore.new(path: 'avro/schemas')
   config.registry_url = Rails.configuration.x.avro_schema_registry_url
-  config.build_schema_registry!
+
```
jturkel, line 97: Can we add a comment that `build_messaging!` will automatically call `build_schema_registry!` if a `schema_registry` hasn't been configured?

jturkel (changes requested):
Thanks for the contribution! Looks like a nice improvement. One small comment about making the relationship clearer between `build_messaging!` and `build_schema_registry!`.

README.md
mkrisher: done, I added an explicit note right under the example, is that what you were thinking?

mkrisher:
I may have additional PRs fwiw, we are looking at potentially using avromatic with debezium events!

jturkel (approved):
Looks great. Thanks again for the contribution. PRs are always welcome for any issues you hit.

PR #153, opened 2023-02-20
fixed issues with bigint 

bartes:
Added ranges for int and bigint according to specs

This fix allows handling situations where `int` and `long` are present and covers issues like:

`Avro::IO::AvroTypeError: The datum 3324284763061284583 is not an example of schema "int"`


bartes:
I will close it for now, because it actually requires some additional testing

PR #159, opened 2023-10-12
Support rails 7.1

kbarrette:
* Support rails 7.1
* Drop support for old ruby and rails
* Drop support for avro 1.10


avromatic.gemspec
```diff
@@ -31,8 +31,8 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.7'
 
-  spec.add_runtime_dependency 'activemodel', '>= 5.2', '< 7.1'
-  spec.add_runtime_dependency 'activesupport', '>= 5.2', '< 7.1'
+  spec.add_runtime_dependency 'activemodel', '>= 5.2', '< 7.2'
+  spec.add_runtime_dependency 'activesupport', '>= 5.2', '< 7.2'
   spec.add_runtime_dependency 'avro', '>= 1.10.0', '< 1.12'
```
jturkel, line 36: Can we drop support for avro 1.10.x if we're making a breaking change anyway?

kbarrette: @jturkel done

avromatic.gemspec
```diff
@@ -31,9 +31,9 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.7'
 
-  spec.add_runtime_dependency 'activemodel', '>= 5.2', '< 7.1'
-  spec.add_runtime_dependency 'activesupport', '>= 5.2', '< 7.1'
-  spec.add_runtime_dependency 'avro', '>= 1.10.0', '< 1.12'
+  spec.add_runtime_dependency 'activemodel', '>= 5.2', '< 7.2'
+  spec.add_runtime_dependency 'activesupport', '>= 5.2', '< 7.2'
+  spec.add_runtime_dependency 'avro', '>= 1.11.0', '< 1.12'
```
jturkel, line 36: Looks like we have an `Avromatic::allow_decimal_logical_type?` helper function and some associated conditionals that can be removed now that we're dropping Avro 1.10 support.

kbarrette: I'll do a deeper investigation into what's avro 1.10-related.

kbarrette: @jturkel as far as I can tell that helper is the only version comparison we're doing here. I have removed it and any uses of it.

jturkel (approved):
🎉

PR #164, opened 2024-09-09
Add support for Rails 7.2

hughevans:
Not sure if you’d prefer to open-end the version check here or use `< 8`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,7 @@ workflows:
                 - gemfiles/avro1_11_rails6_1.gemfile
                 - gemfiles/avro1_11_rails7_0.gemfile
                 - gemfiles/avro1_11_rails7_1.gemfile
+                - gemfiles/avro1_11_rails7_2.gemfile
               ruby-version:
                 - 3.0.6
                 - 3.1.4
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,3 +78,6 @@ workflows:
                 - 3.1.4
                 - 3.2.2
                 - 3.3.0
+            exclude:
+              - gemfile: gemfiles/avro1_11_rails7_2.gemfile
+                ruby-version: 3.0.6
--- a/Appraisals
+++ b/Appraisals
@@ -17,3 +17,9 @@ appraise 'avro1_11-rails7_1' do
   gem 'activesupport', '~> 7.1.1'
   gem 'activemodel', '~> 7.1.1'
 end
+
+appraise 'avro1_11-rails7_2' do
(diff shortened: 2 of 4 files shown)


Appraisals
```diff
@@ -17,3 +17,9 @@ appraise 'avro1_11-rails7_1' do
   gem 'activesupport', '~> 7.1.1'
   gem 'activemodel', '~> 7.1.1'
 end
+
+appraise 'avro1_11-rails7_2' do
```
jturkel, line 21: Can you add this new gemfile to the build [matrix](https://github.com/salsify/avromatic/blob/master/.circleci/config.yml#L69)?

avromatic.gemspec
```diff
@@ -31,8 +31,8 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.7'
 
-  spec.add_runtime_dependency 'activemodel', '>= 5.2', '< 7.2'
-  spec.add_runtime_dependency 'activesupport', '>= 5.2', '< 7.2'
+  spec.add_runtime_dependency 'activemodel', '>= 5.2'
```
jturkel, line 34: Let's go with a version constraint of `< 8` so we can ensure compatibility with major rails releases.

jturkel (changes requested):
Thanks for the contribution @hughevans! A few things to change but then we should be able to get this released.

Appraisals
hughevans: Not a problem, done 👍

avromatic.gemspec
hughevans: No worries 👍

.circleci/config.yml
```diff
@@ -77,3 +77,11 @@ workflows:
                 - 3.1.4
                 - 3.2.2
                 - 3.3.0
+      - test_modern:
```
jturkel, line 80: CI is failing because the `test_modern` job isn't defined. Can you instead add the `gemfiles/avro1_11_rails7_2.gemfile` Gemfile to the above build matrix and use appropriate [exclusions](https://circleci.com/docs/configuration-reference/#excluding-sets-of-parameters-from-a-matrix) for unsupported versions of Ruby? Here are some [examples](https://github.com/salsify/goldiloader/blob/master/.circleci/config.yml#L92) from other projects.

jturkel (changes requested):

.circleci/config.yml
hughevans: Ah gotcha. Thanks for explaining that.

jturkel (approved):
Thanks for the contribution @hughevans. I'll get a release out shortly.

PR #166, opened 2024-11-05
Improve support for BigInt, enhance union type

opti:
- Add ranges for int and bigint according to specs.
    This fix allows handling situations where int and long are present and covers issues like:
    ```ruby
    Avro::IO::AvroTypeError: The datum 3324284763061284583 is not an example of schema "int"
    ```
- Enhance union type selection during the serialization. Previously, we only checked the value class (now additionally for the union types index picking check if the value is coercible with a particular type)


opti:
Not sure why Linter is complaining. There is no changes in `lib/avromatic/model/raw_serialization.rb:19:9` in this PR.
```
Inspecting 94 files
.......................C......................................................................

Offenses:

lib/avromatic/model/raw_serialization.rb:19:9: C: Style/AccessModifierDeclarations: private should not be inlined in method definitions.
        private :datum_writer, :datum_reader
        ^^^^^^^

94 files inspected, 1 offense detected

Exited with code exit status 1
```

erikkessler1:
> Not sure why Linter is complaining. There is no changes in lib/avromatic/model/raw_serialization.rb:19:9 in this PR.

I'd guess a new version of RuboCop got used that changed how it it applies that rule. You could add `# rubocop:disable Style/AccessModifierDeclarations` to that line since it seems like a false positive.

lib/avromatic/model/types/union_type.rb
```diff
@@ -76,9 +76,8 @@ def referenced_model_classes
         private
 
         def find_index(value)
-          # TODO: Cache this?
```
erikkessler1, line 79: This can be a hot code path, so it would be nice to be able to cache this index lookup (i.e., build a cache from `value.class` to `union_index`). With this change, since we match on the `value` itself, that cache isn't possible.

Additionally, the `matched?` call adds overhead for all types, but we only need it for the case where `value.is_a?(Integer)`. What do you think about special casing integer handling here?

```ruby
def find_index(value)
  if value.is_a?(::Integer)
    if Avromatic::Model::Types::IntegerType.in_range?(value)
      member_types.index(Avromatic::Model::Types::IntegerType)
    end || if Avromatic::Model::Types::BigIntType.in_range?(value)
      member_types.index(Avromatic::Model::Types::BigIntType)
    end
  else
    # TODO: Cache this
    member_types.find_index do |member_type|
      member_type.value_classes.any? { |value_class| value.is_a?(value_class) }
    end
  end
end
```

erikkessler1 (commented):
Thank you for working on this! It certainly seems like a gap.

A question I have is if we can remove the `#matched?` in favor of special casing integer handling in `UnionType#find_index` since we really only need `matched?` for integers.

lib/avromatic/model/types/union_type.rb
opti: @erikkessler1 fair enough. I've added special casing for Integer.

opti: P.S. is there anything different for this CI run?

The same specs locally are running with no issues:

```
❯ bundle exec rspec --format progress spec
...............................................................................................................................................................................................................................................................................................................................................

Finished in 0.53541 seconds (files took 0.60385 seconds to load)
335 examples, 0 failures

Coverage report generated for RSpec to avromatic/coverage.
Line Coverage: 96.59% (1245 / 1289)
```

erikkessler1: It seems like the 403 is coming from Sinatra ([documentation](https://sinatrarb.com/intro.html#:~:text=if%20not%20available.-,host_authorization,-You%20can%20pass)) as the host doesn't match the `permitted_hosts`. This must have changed between Sinatra versions. I was able to get things passing locally by changing the `Avromatic.registry_url` [here](https://github.com/salsify/avromatic/blob/5899c5e7efdd7d2dec6b5b63951c2ea3cf831de4/spec/spec_helper.rb#L22):

```diff
-Avromatic.registry_url = 'http://username:password@registry.example.com'
+Avromatic.registry_url = 'http://username:password@registry.localhost'
```

lib/avromatic/model/types/abstract_type.rb
```diff
@@ -23,12 +24,16 @@ def coerce(_input)
...
+          input.nil? || input_classes.any? { |input_class| input.is_a?(input_class) }
         end
 
-        def coerced?(_value)
-          raise "#{__method__} must be overridden by #{self.class.name}"
+        def coerced?(value)
+          value.nil? || value_classes.any? { |value_class| value.is_a?(value_class) }
+        end
+
+        def matched?(value)
```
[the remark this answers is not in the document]
opti: Right. Removed.

lib/avromatic/model/types/union_type.rb
opti: Updated.

spec/avro/schema/test/real_int_union.avsc
erikkessler1, line 1: For completeness, we probably want the DSL that generated this file similar to https://github.com/salsify/avromatic/blob/master/spec/avro/dsl/test/real_union.rb

erikkessler1 (commented):
One final thought on the spec fixture you added that I missed earlier. Looks great otherwise!

spec/avro/schema/test/real_int_union.avsc
opti: Added. I've run the `avro:generate_spec` to double check as well.

There was one missing logical type in generated schemas, so added that as well.

erikkessler1 (approved):
Awesome! Thanks for your work on this. I'll cut a new version with these changes.

erikkessler1:
Version 5.1.1 should now be available

PR #171, opened 2025-10-16
Avro 1.12 Support

jturkel:
This PR adds support for Avro 1.12. It also drops support for EOL versions of Ruby/Rails and loosens the Rails dependency since we're only using a few stable public APIs.

@atsheehan - you're prime


atsheehan (approved):

.ruby-version
```diff
@@ -1 +1 @@
-3.3.5
+3.2.9
```
marcoserrato, line 1: Curious about this version downgrade?

jturkel: For gems this is just the default version when running the project locally. It won't impact the released gem. We generally default to using the oldest supported version.

